PR #5, opened 2015-06-08
Feature: Multiline comments, strings, etc

ptersilie:
This patch adds multiline capabilities to Eco which enable comments and strings containing newlines.


ptersilie:
Moved to #20.

PR #7, opened 2015-10-02
Tidy up Python in TUTORIAL.

snim2:
Changes a variable name and corrects indentation in tutorial instructions.
Tidy up output so HTML just contains names and not the tuple repr (e.g. `<li> Zara Mitchell` rather than `<li> (u'Zara Mitchell',)`).

### DIFF
--- a/TUTORIAL
+++ b/TUTORIAL
@@ -66,7 +66,7 @@ Move you cursor right after "<body>" hit ENTER and open up the language menu
 again by hitting CTRL+L. Again select "Python + HTML + SQL" from the list and
 then type:
 
-    for n in db:
+    for n, in db:
         print html, n
 
 In this code "db" and "html" are placeholders that we are going to replace with
@@ -92,7 +92,7 @@ Your code now should look somewhat like this:
     </head>
     <body>
     for n, in SELECT name FROM people:
-    print <li>, n
+        print <li>, n
     </body>
     </html>
 


ltratt:
Does the `SELECT` return a tuple of size 1? I had assumed (though my memory is bad!) that it returned the actual element.

snim2:
`SELECT` always returns a tuple, IIRC.

This is an example from the `sqlite3` docs:

``` python
>>> for row in c.execute('SELECT * FROM stocks ORDER BY price'):
        print row

(u'2006-01-05', u'BUY', u'RHAT', 100, 35.14)
(u'2006-03-28', u'BUY', u'IBM', 1000, 45.0)
(u'2006-04-06', u'SELL', u'IBM', 500, 53.0)
(u'2006-04-05', u'BUY', u'MSFT', 1000, 72.0)
```

ptersilie:
Tuples is what I also would have expected.

ltratt:
Clearly if we select multiple column it will return a tuple (or similar compound element). I'm asking what happens if you select a _single_ element (as in the example).

snim2:
The Python library isn't smart enough to know that a single column is being selected, e.g.:

``` python
>>> import sqlite3
>>> conn = sqlite3.connect('example.db')
>>> c = conn.cursor()
>>> c.execute("""CREATE TABLE Example
... (value text)""")
<sqlite3.Cursor object at 0x7f6edd810b90>
>>> c.execute("""INSERT INTO Example VALUES ('hello')""")
<sqlite3.Cursor object at 0x7f6edd810b90>
>>> c.execute("""INSERT INTO Example VALUES ('goodbye')""")
<sqlite3.Cursor object at 0x7f6edd810b90>
>>> for val in c.execute("""SELECT value text FROM Example"""):
...     print val, type(val)
... 
(u'hello',) <type 'tuple'>
(u'goodbye',) <type 'tuple'>
>>> 

```

ltratt:
Ah, yes, I remember that when I do this example I write:

``` python
for name, in SELECT name FROM table:
   ...
```

Which I think is a bit nicer than `result[0]`, but I don't feel strongly about it either way.

snim2:
Right, I see what happened here -- the comma appears in one part of the TUTORIAL and not another. I'll re-write this PR and rebase it...

snim2:
Done -- `n` is back and the tuple is no longer subscripted.

ltratt:
Looks good.

PR #11, opened 2015-10-05
Load .ui files directly. 

snim2:
This PR removes all the auto-generated code which defines the GUI. 

Instead, each `.ui` file is loaded directly. This means the `.py` files in the `lib/eco/gui` no longer need to be kept in sync with the `.ui` files.

Note that this is the smallest possible change that removes the auto-generated code. It would be possible to replace the `self.ui` instance variables in each class with inheritance. For example:

``` python
class AboutView(QtGui.QDialog):
    def __init__(self, parent):
        self.parent = parent
        QtGui.QDialog.__init__(self, parent)
        self.ui = Ui_About()
        self.ui.setupUi(self)
```

would become:

``` python
class AboutView(QtGui.QDialog, Ui_About):
    def __init__(self, parent):
        self.parent = parent
        QtGui.QDialog.__init__(self, parent)
        self.setupUi(self)
```

but this would touch many more lines of code in `eco.py`.


ltratt:
I like the idea of this! Though it's for @ptersilie to have the final say.

ptersilie:
Like it. :+1: Will test and merge tomorrow.

snim2:
Grand. Do say if you want me to neaten this up a bit more. Potentially the `self.parent = parent` lines could go.

ptersilie:
Sure. If you want to make the code more concise I won't object. :-)

ptersilie:
Newfile dialog is not working. Needs the following patch:

```
diff --git a/lib/eco/eco.py b/lib/eco/eco.py
index 347b32f..9bfae4a 100644
--- a/lib/eco/eco.py
+++ b/lib/eco/eco.py
@@ -766,7 +766,7 @@ class Window(QtGui.QMainWindow):

     def newfile(self):
         # create new nodeeditor
-        lview = LanguageView(languages)
+        lview = LanguageView(self, languages)
         result = lview.exec_()
         if result:
             etab = EditorTab()

```

ltratt:
This suggests that it would be best to do this as a branch of the main eco repository (so Lukas can push that commit), rather than on a fork.

ltratt:
If you guys set that up, then raise another pull request, I'll do a review and (assuming all's good) merge.

ptersilie:
Definitely. I'm always for branches in the original repo if the authors are part of the team. Otherwise I have to find the correct repo on GitHub and clone it first before I can test it. And of course I can't commit to it.

snim2:
OK, I have fixed the call to the `LanguageView` constructor on line 769 and rebased, so the history is clean.

In future, if working on a branch of the main repo is what you normally to, I'll fall into step. However, working on a fork prevents anyone from causing chaos if they accidentally push to master. You can locally checkout and make changes, including adding new commits, to any pull request in your own GitHub repos by following the instructions here: https://help.github.com/articles/checking-out-pull-requests-locally/ -- there shouldn't really be a need to clone the fork, AFAICT.

ptersilie:
Ah that makes things a lot easier. I still believe that anyone who is part of the project (i.e. not an external) should just create branches on the main repo, but I'm not too fussed either way.

snim2:
No problem, if that is the normal pattern I'll do that in future.

ltratt:
Good point about pulling in things. I think I'm OK either way, at least for the time being. I'd suggest: keep on working in a fork. If we find it becomes a pain, we can change our process later.

ptersilie:
Sure. I'm happy with either of the solutions. Now that I know how to pull in things just do whatever works best for you Sarah. :)

PR #16, opened 2015-10-16
Add profiling to Python2 code. 

snim2:
Adds a menu item for the profiler. Callls a profiler from the Python2 stdlib and adds anotations to each function definition. Annotations are bold until the code is 'stale' -- i.e. it has been modified.

Things that need to be done before this PR can be merged:
- ~~Disable all profiling / tooling actions when the language of the current language does not support profiling~~
- ~~Run profilers and other tools in a worker thread (helpful when editing long-running programs)~~ (worker thread in `eco.py`)
- ~~Display a throbber while collecting profiling information (helpful when editing long-running programs)~~
- ~~When profiling, any output to STDOUT should appear in the Eco console~~


snim2:
I'm not sure about this. The annotation is longer than the node above it, and with your method the annotation is written under the "def" but is cut off after three characters.

snim2:
OK, the bug mentioned in the comment above is now fixed.

snim2:
OK, I think this is ready for review...

lib/eco/treemanager.py
```diff
@@ -526,6 +543,18 @@ def find_next(self):
         if self.last_search != "":
             self.find_text(self.last_search)
 
+    def find_text_no_cursor(self, text, parent_name='funcdef'):
```
ptersilie, line 546: Is this still used?

snim2, line 546: Good point. I'm not using it in this PR but it is likely to be used later on.

ptersilie, line 546: Ok.

ptersilie:
I think you forgot to add throbber.gif.

snim2:
Added the throbber, when you have finished reviewing I'll squash that commit.

lib/eco/eco.py
```diff
@@ -1042,5 +1064,49 @@ def run(self):
...
+        self.tab_bar = tab_bar
+        self.setAlignment(Qt.AlignCenter)
+        self._movie = QMovie("gui/throbber.gif")
+        self.setMovie(self._movie)
+
+    def hide(self):
+        """Hide throbber in left hand corner of tabbed notebook.
+        """
+        self._movie.stop()
+        self.tab_bar.setCornerWidget(None, corner=Qt.TopLeftCorner)
```
ptersilie: It doesn't matter a great deal, but I think the waiting animation (which I quite like by the way) looks a bit better on the top right corner. What do you think?

ptersilie:
Awesome, thanks. Rest looks good. If you squash I will merge. :)

snim2:
Great, squashed!

PR #17, opened 2015-10-21
Convert newlines when copying text from eco

ptersilie:
Newlines in Eco are represented as `\r`. When copying text from within Eco and pasting it into another text editor (e.g. vim) the newlines are no recognized properly. This patch fixes it by converting copied text to `\n`.


snim2:
lgtm

PR #22, opened 2015-11-10
Export or run a SimpleLanguage program.

snim2:
Here is a [simple example](https://youtu.be/1KiCJSFz9co)

### DIFF
--- a/lib/eco/treemanager.py
+++ b/lib/eco/treemanager.py
@@ -1427,13 +1427,39 @@ class TreeManager(object):
             return self.export_php_python(path, run)
         elif lang == "Python 2.7.5":
             import tempfile
-            import os, sys, subprocess
+            import subprocess
             f = tempfile.mkstemp()
             self.export_as_text(f[1])
             print "python " + f[1]
             return subprocess.Popen(["python2", f[1]], stdout=subprocess.PIPE, stderr=subprocess.STDOUT, bufsize=0)
+        elif lang == "SimpleLanguage":
+            return self.export_simple_language(path, run)
         else:
             return self.export_as_text(path)
+
+    def export_simple_language(self, path=None, run=False):
+        import os
+        import os.path
+        import tempfile
+        import subprocess
+        import sys
+        if run:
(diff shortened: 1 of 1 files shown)


lib/eco/treemanager.py
```diff
@@ -1427,14 +1427,40 @@ def export(self, path=None, run=False):
...
+            if not os.environ.has_key("GRAAL_WORKSPACE"):
+                sys.stderr.write("GRAAL_WORKSPACE environment not set")
+                return
+            working_dir = os.path.join(os.environ["GRAAL_WORKSPACE"], "graal-compiler")
+            f = tempfile.mkstemp(suffix=".sl")
+            self.export_as_text(f[1])
+            # Run this command:
+            #     $ cd $GRAAL_WORKSPACE/graal-compiler
+            #     $ ../../mx/mx --vm graal sl sl_hello_world.sl
+            return subprocess.Popen(["../../mx/mx", "--vm", "graal", "sl", f[1]],
```
ltratt, line 1456: Previous (bitter) experience suggests we should be passing MX in with an environment variable (or similar).

snim2, line 1456: `$GRAAL_WORKSPACE` should be the whole workspace which holds all the compilers and tools, so it should contain `mx`.

ltratt, line 1456: Right, but the `../../mx/mx` path isn't relative to `$GRAAL_WORKSPACE_ in the`Popen` call.

ltratt, line 1456: I'm wrong! Because of `cwd`.

snim2, line 1456: No, the `Popen()` call is passed a `cwd` argument which sets its working directory.

ltratt, line 1456: Yep, I'd spotted I'd got that wrong. So ignore me :)

ptersilie:
Looks good. I like the youtube videos. :) And that I can hear your pc fans speeding up when you run the second example. ;)

PR #23, opened 2015-11-11
Add support for profiling SimpleLanguage coding using the SL coverage counter

snim2:
Example demo [can be found here](https://youtu.be/ZhHXNJDf_po)

### DIFF
--- a/lib/eco/nodeeditor.py
+++ b/lib/eco/nodeeditor.py
@@ -369,10 +369,11 @@ class NodeEditor(QFrame):
                 else:
                     self.infofont.setBold(False)
                 paint.setFont(self.infofont)
+                paint.setPen(QPen(QColor((highlighter.get_default_color()))))
                 start_x = (0 if (x - len(prof) * self.infofontwt) < 0
                              else x - len(prof) * self.infofontwt)
                 start_y = self.fontht + ((y + 1) * self.fontht)
-                paint.drawText(QtCore.QPointF(start_x, start_y), prof)
+                paint.drawText(QtCore.QPointF(x-dx, start_y), prof)
                 self.lines[line].height = max(self.lines[line].height, 2)
                 paint.setFont(self.font)
 
--- a/lib/eco/syntaxhighlighter.py
+++ b/lib/eco/syntaxhighlighter.py
@@ -67,6 +67,9 @@ class SyntaxHighlighter(object):
             hexcode = self.colors[color]
         return hexcode
 
+    def get_default_color(self):
+        return self.palette.color(QPalette.Text)
+
     def get_style(self, node):
         if node.symbol.name in self.keyword_style:
             return self.keyword_style[node.symbol.name]
--- a/lib/eco/treemanager.py
+++ b/lib/eco/treemanager.py
@@ -298,7 +298,7 @@ class TreeManager(object):
             "PHP + Python" : False,
             "PHP" : False,
             "Python 2.7.5" : True,
-            "SimpleLanguage" : False,
+            "SimpleLanguage" : True,
         }
         self.input_log = []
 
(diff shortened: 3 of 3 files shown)


ptersilie:
Why are some of the profiling lines indented and others are not? I wonder if we should put the profiling results actually at the end of the line it is profiling.

snim2:
Hmm, good catch. Looking through the text output of the coverage counter I can't see anything obvious. Watch this space....

ptersilie:
It might be this line:

```
temp_cursor.move_to_x(len(msg) *  30, self.lines)
```

I don't know what the magic number 30 does. ;)

snim2:
OK, pushed a fix and the result looks like this:

sl-profile-screenie

ptersilie:
If you want you can you try

```
temp_cursor.right()
node = temp_cursor.find_next_visible(temp_cursor.node)
if node.lookup == "<ws>":
    node = node.next_term
self.profile_map[node] = msg
```

That should align the profiling information with the indentation of the line it profiles.

snim2:
That sounds like a good idea, but I can't get it to display more than one line of profile information.

If I do something like this:

``` python
                    temp_cursor.line = lineno - 1
                    temp_cursor.move_to_x(0, self.lines)
#                    temp_cursor.right()
                    node = temp_cursor.find_next_visible(temp_cursor.node)
                    if node.lookup == "<ws>":
                        print("Line", lineno, "moving passed WhiteSpace")
                        node = node.next_term
                    self.profile_map[node] = msg
```

then the output of those print statements looks like this:

```
('Line', 2, 'moving passed WhiteSpace')
('Line', 3, 'moving passed WhiteSpace')
('Line', 4, 'moving passed WhiteSpace')
('Line', 5, 'moving passed WhiteSpace')
('Line', 6, 'moving passed WhiteSpace')
('Line', 8, 'moving passed WhiteSpace')
('Line', 12, 'moving passed WhiteSpace')
('Line', 13, 'moving passed WhiteSpace')
('Line', 14, 'moving passed WhiteSpace')
('Line', 15, 'moving passed WhiteSpace')
('Line', 16, 'moving passed WhiteSpace')
('Line', 17, 'moving passed WhiteSpace')
('Line', 18, 'moving passed WhiteSpace')
```

and the profile info still prints in the same place, but with the foreground colour of the leftmost lexeme on the line above.

I think this because our two bits of code are not playing nicely with each other. In `nodeeditor.py` we have:

``` python
            # Draw profiling information.
            if node in self.tm.profile_map:
                prof = self.tm.profile_map[node]
                if not self.tm.profile_is_dirty:
                    self.infofont.setBold(True)
                else:
                    self.infofont.setBold(False)
                paint.setFont(self.infofont)
                start_x = (0 if (x - len(prof) * self.infofontwt) < 0
                             else x - len(prof) * self.infofontwt)
                start_y = self.fontht + ((y + 1) * self.fontht)
                paint.drawText(QtCore.QPointF(start_x, start_y), prof)
                self.lines[line].height = max(self.lines[line].height, 2)
                paint.setFont(self.font)
```

ptersilie:
Hmm, unfortunately I can't test this as I don't have GRAAL installed on my machine. It would be interesting to know if the new code moves to the right node and it's indeed a problem in nodeeditor.py.

How can I get Graal running to test this? Is it a lot of work?

snim2:
It is a lot of work, the easier thing to do would be to see whether your snippet works with the cPython profiler.

ltratt:
@snim2, ISTR you have a build script that automates building graal (based on Edd's stuff)? Not that it's an immediate concern, but it might be worth giving that to Lukas at some point, if it's practical. [Since Graal won't build on OpenBSD, I can't run it yet...]

snim2:
It isn't quite so simple because I have some patches and so on that aren't in the public repos. This is something we should probably discuss in detail in the sprint, as it's the sort of thing that is likely to get more rather than less difficult as time goes on.

ltratt:
Agreed on all points.

snim2:
OK, I think this is all fixed now. Example SL profile session:

slprofiler-indented

and an example Python profiling session:

pyprofiler-indented

ltratt:
Looks nice to me. I wonder if we can do a heatmap from this though? The only problem with a huge program is going to be finding the needle in a haystack if you have to look at lots of tiny bits of information.

snim2:
A heatmap is the next thing on the list. but I think that should be a separate PR.

I think the concrete information is also useful though, and if the user wants to see language boxes rather than the heatmap, having the text is better than nothing.

ltratt:
Sounds like a sensible plan to me.

ptersilie:
@snim2 Is this ready to merge then?

snim2:
Yes, if you are happy and you don't want me to squash any commits?

ptersilie:
Looks good to me. :)

Issue #6, opened 2015-08-07
Make current language box more obvious


vext01:
In the presence of many language boxes it can be really hard to know which box you are in. Although the boxes are different colours:
- The colours can be very similar, making telling them apart hard.
- The colours of the boxes change depending where the cursor is. I find it hard to build a mental map of the program I am editing when it keeps changing appearance.

We discussed putting a dashed border around the current box, whilst also colouring other boxes, but there may be issues with this.

Situation where colours are too subtle:
https://farm1.staticflickr.com/378/19751307154_2387137aa5_o.png

Example of dashed box idea:
https://farm4.staticflickr.com/3720/19753494503_2a8db22524_o.png

Hope this helps, Thanks

snim2:
What was the issue with putting a border around the language boxes? It sounds like a good idea to me.

ltratt:
If you have boxes in boxes in boxes, you'd end up with lots of borders.

That said, some users may like this (even though I'd probably find it too cluttered).

ptersilie:
The dashed line is currently hard to do code-wise, as it is not clear when drawing the source code where it starts and ends. I can have a look at it but I already know that the solution would be quite complex and not worth the trouble, given that we are gonna move onto Eco 2 in the future.
Since dashes are hard I colored the active language box differently instead, which I agree can be a bit confusing.

I agree, that we have to think carefully about fixed colors and how we can improve on the visuals for language boxes.

vext01:
Many thanks!

PR #25, opened 2015-11-17
Show brackets around currently selected language box

ptersilie:


ptersilie:
Fixes #6

lib/eco/nodeeditor.py
```diff
@@ -446,6 +448,34 @@ def paint_nodes(self, paint, node, x, y, line, max_y, lbox=0):
 
         return x, y, line
 
+    def draw_lbox_bracket(self, paint, bracket, node, x, y):
```
ltratt, line 485: I think we should add a line along the lines of `assert bracket in ["[", "]"]` to avert potential misuses of the `bracket` variable (which otherwise won't be caught and will lead to confusion).

ptersilie:
Fixed and rebased.

ltratt:
Good stuff.

PR #29, opened 2015-11-18
Highlight comments grey

### LINKED ISSUE #28: Highlight comments properly
With the merge of #20, comments are not highlighted properly any more, as the highlighter only works on tokens. For the new comments, we need to make it work on nonterminals including their children.

ptersilie:
Since the highlighter uses the parse tree to determine the colors,
this only works if the program was succesfully parsed. As long
as there is a parsing error, newly added comments won't be highlighted
correctly.

Fixes #28


ltratt:
Apart from the "u" on your keyboard apparently being broken, this looks good.

PR #32, opened 2015-11-19
Improved mouse selection

ptersilie:
This fixes several issues with selection text in Eco. Text can now be selected past the currently visible lines. Selection text using the mouse now automatically scrolls the text when the mouse reaches the borders of the textarea.

/cc @ltratt


snim2:
Works for me!

PR #20, opened 2015-11-10
Multiline and indentation support inside the parser

ptersilie:
Under development

PR #43, opened 2015-11-23
Use multiline strings in Java

ptersilie:


ltratt:
Given that the diff isn't super useful, could you maybe provide a super quick explanation of what the changes to the Eco grammar file are? A sentence or two should be fine.

ptersilie:
I changed Java to make use of the multiline strings that were merged in #20

ltratt:
OK, makes sense.

PR #46, opened 2015-11-24
Fix arrow keys at start / end of lines.

### LINKED ISSUE #45: Fix arrow keys in eco.treemanager.cursor class
If the cursor is on the 0th column of a line, pressing the left arrow key should move the cursor to the right-hand side of the right-most node on the line above.

If the cursor is on the right-hand side of the right-most node on a line, pressing the right arrow key should move the cursor to the 0th column of the line below.

snim2:
Fixes #45

### DIFF
--- a/lib/eco/treemanager.py
+++ b/lib/eco/treemanager.py
@@ -89,7 +89,7 @@ class Cursor(object):
         if not self.is_visible(node):
             node = self.find_previous_visible(self.node)
         if node.symbol.name == "\r":
-            return
+            self.line -= 1
         if isinstance(node, BOS):
             return
         if not node is self.node:
@@ -116,7 +116,7 @@ class Cursor(object):
         else:
             node = self.find_next_visible(node)
             if node.symbol.name == "\r":
-                return
+                self.line += 1
             if isinstance(node, EOS):
                 return
             self.node = node


ltratt:
Fixing this would be nice. The solution for "left" looks sensible to me; I have no real clue whether the "right" solution is the best one or not though. So I'm assigning Lukas :)

ptersilie:
I never thought I'd say this but, it's not a bug, it's a feature. Haha. In terms of behaviour in Eco, I always went with whatever the only editor I have does, Vim. But I'm not fixed on this behaviour. So if you think it's sensible to change this then we can do.

Let me just have a quick look at the changes, as Ecos default behaviour would automatically jump to the previous line and I changed it so it stops at newlines. This means we should probably remove code and not add more. But let me double check that quickly.

snim2:
Interesting, every editor I have ever used has moved to the previous / next line!

snim2:
Let me know if you want to change the code in `right()` before squashing....

snim2:
OK, this works fine for me, and it does what sublime and emacs do by default.

ptersilie:
Looks good now. Once it's squashed I can merge.

snim2:
Rebased...

snim2:
Rebased and squashed...

snim2:
Thanks!

PR #47, opened 2015-11-24
Fix leftover characters in Python

ptersilie:
Fix Python to lex all characters that are not part of any other lexing rules.


ltratt:
This seems to fix my problem. Good stuff!

PR #51, opened 2015-11-24
Fix language box brackets in column 0

ptersilie:
Brackets showing language boxes were cut off. Moved them slightly
to the right if the box starts at the beginning of the line.

### DIFF
--- a/lib/eco/nodeeditor.py
+++ b/lib/eco/nodeeditor.py
@@ -72,7 +72,7 @@ class NodeEditor(QFrame):
         self.connect(self.blinktimer, SIGNAL("timeout()"), self.trigger_blinktimer)
         self.show_cursor = True
 
-        self.boxcolors = [QColor("#DC322F"), QColor("#268BD2"), QColor("#D33682"), QColor("#B58900"), QColor("#2AA198"), QColor("#859900")]
+        self.boxcolors = [QColor("#DC322F"), QColor("#268BD2"), QColor("#2Ac598"), QColor("#D33682"), QColor("#B58900"), QColor("#859900")]
         self.setCursor(Qt.IBeamCursor)
 
         # Semi-transparent overlay.
@@ -288,7 +288,7 @@ class NodeEditor(QFrame):
         elif settings.value("app_theme", "Light").toString() in ["Gruvbox"]:
             alpha = 100
         else:
-            alpha = 40
+            alpha = 60
 
         first_node = node
         selected_language = self.tm.mainroot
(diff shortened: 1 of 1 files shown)


ltratt:
Yes, I'd noticed this. Excellent!

ltratt:
Ah, but this causes me a problem as now it obscures the first letter inside the language box...
overlap

ptersilie:
Ye, I know, but there is nothing I can do about that (for now). So it's either half the bracket or a bit of the first letter missing. I went for the latter option.

ptersilie:
It's also less obvious with bigger font sizes.

ltratt:
If the vertical stroke of the bracket was only one pixel wide, that would also help?

ptersilie:
Yes, that would help slightly, but brackets are not as visible if we do that. I played around with the alternatives and found the current one the least annoying. But I guess in the end it comes down to personal preference. I'm happy for any other/better solution.

Note though that this only ever is a problem when the language box is at the very beginning of the line.

ltratt:
Maybe we could always have a 1px border around the edges of the file or something like that?

ptersilie:
You mean around the box?

ltratt:
Around the drawn window (i.e. there would be a 1px border where text wouldn't be drawn, but the language box brackets could be).

ptersilie:
Ye I thought about doing this eventually. But that involves touching a load of files to make up for that border again. :/

ltratt:
OK... I must admit, I find this PR a bit hard to know what to do with. It looks really bad when the text is overwritten :( But I can see that a better fix might be harder.

One possibility might be to draw the bracket under the text in a lighter / opaque way? Or maybe draw the brackets _until_ the user types something then make them disappear?

snim2:
One quick suggestion -- how about making the space between each line slightly bigger, then the top and bottom of the brackets will be more visible, even if the vertical isn't?

ltratt:
Ah, yes, that might be a cunning interim hack.

ptersilie:
It would still look confusing if the left bracket is cut off while the right one isn't. Also increasing the y-height has the same issues as moving the x-axis. It destroys things like mouse clicking for instance that need to be adjusted for the altered x,y positions. And I don't know at the moment what else is affected. This is why this is a bit of a bigger project to fix properly.

At the moment it seems we have two options: we either use this fix, or we remove the brackets for now until I have time to fix this properly.

ltratt:
It seems to me that the compromise option is still to make the vertical stroke 1px wide. I know it's not quite as obvious what it's doing, but at least it won't obscure text. If you guys are OK with that, I think I'd be happy to merge such a thing.

vext01:
I've noticed this also. FWIW: you guys have already discussed all of the ideas I had to potentially fix this.

ptersilie:
@vext01 I fixed this issue and updated the languagebox colors and the color of the brackets as well. Have look and see if you like it.

@ltratt If Edd is happy, this can be merged.

ltratt:
I like it. Edd: let me know if you're OK and then I'll merge.

vext01:
I've had a play around and I think this is good.

PR #52, opened 2015-11-24
Export images properly when exporting Html+Python

ptersilie:
Images weren't exported previously and resulted in an error. Given that the grammars allow to include them we should export them too.


ltratt:
Good point!

PR #53, opened 2015-11-24
Bring truffle code up to date with upstream repo

snim2:


lib/eco/export/jruby_simple_language.py
```diff
@@ -92,24 +92,18 @@ def _walk_rb(self, node):
...
 
-    def _apply_template(self, name, magic_name):
-        return """  {1} = Truffle::Interop.import('{0}')
-  def {0}(*args)
-    Truffle::Interop.execute({1}, *args)
-  end
-
-""".format(name, magic_name)
+    def _apply_template(self, name):
+        return "Truffle::Interop.import_method(:{0})".format(name)
```
ltratt: Is there any reason for using `format` over `%`?

snim2: Nothing in particular

ltratt: Given that `%` is more common, should we use that then?

ltratt:
Do you want to squash, then we can merge?

snim2:
OK, all squashed...

ltratt:
Thanks!

ptersilie:
Even though I prefer the `%` myself, it seems `format` is the way to go: https://docs.python.org/release/3.1.5/library/stdtypes.html#old-string-formatting-operations

> Note: The formatting operations described here are obsolete and may go away in future versions of Python. Use the new String Formatting in new code.

ltratt:
I don't mind too much, but I think we should be consistent within this codebase. Currently we use `%`. We could easily replace all the `%`s with `format` in a PR, and then we could move over to `format` permanently. [I'm not going to raise such a PR, or be hugely interested in reviewing it, but I wouldn't object to it either.]

snim2:
OK, I will add this as an issue so we don't forget, but I guess it won't be a priority any time soon.

ptersilie:
My point wasn't that we should switch over to `format`. I just wanted to point out that it's ridiculous that we are even discussing whether we should use one or the other, because it is completely irrelevant and not worth our time.

ltratt:
Code consistency within a codebase isn't irrelevant -- the less consistent we are, the harder it is to maintain things over time.

PR #56, opened 2015-11-25
Eco plugin API for external tools

### LINKED ISSUE #35: Add a plugin API for Eco
At present, there are a small number of external tools whose data is visualised in Eco. Each one is hard-coded. 

In the future it would be preferable to add new tools without having to interact with the other Eco classes. It would also be nice for the user to be able to choose whether or not to use each available tool with each document they have open.

To achieve this, a plugin API should be created, to allow tool vendors to add Eco support for their tools. This should use the `eco.incparser.annotation` classes and as little other code as possible.

snim2:
This PR provides a plugin API for Eco. The aim is to allow users to annotate
their code with information from external tools.

Advantages of this approach are:
- External tools no longer piggyback on the export / run code, which leaves
  that code cleaner and obviates the need for the `mock` objects which
  were a work-around to make the console work with the CPython profiler.
- Plugins are stored in `~/.eco/` which means that different users can have
  their own tools set up in different ways, according to their own tastes.
- File paths to external tools, which are likely to differ between users,
  are now stored externally in the plugins, and not hard-coded into Eco.
  The exception to this is the **Project -> Run** command which this PR
  does not touch.
- Different users can configure the same tools to produce different
  visualisations (by using _Hints_ differently).
- The **Tools** menu is generated at run-time, according to which plugins
  are loaded.
- Plugins are just ordinary Python code, which should conform to the
  documented API.
- External tools run on a timer and in a separate thread to the main GUI.
- The throbber spins when the user runs a program, not when the external
  tools are running.

In order to test this PR you will need to:
- Fetch and checkout the `eco-plugins` branch
- Create a directory called `~/.eco/`
- Copy the files in `plugins/` into your `~/.eco/` directory
- Open Eco. Check that the **Tools** menu corresponds to your plugins.
  There should be sub-menus for each language which has a plugin.
- The Python plugins should work for everyone and only depend on stdlib.
- Open or create a Python file. You should see tooltips and a heatmap.
- Try turning the Python plugins on and off.

Some subtleties to look out for:
- When more than one document is open in Eco, do the plugins behave
  correctly when you change tabs?
- By design, if a plugin fails after it has been loaded, nothing happens.
  A plugin is a proxy for an external tool, so if one errors, it should not
  interrupt the user. It might be worth making a plugin fail somehow to
  check this.
- Will the right plugins be loaded for the right languages? At the moment
  the `self.language` field of each plugin (which correspond to sub-menus
  under the **Tools** menu) must exactly match the language of the currently
  open document in order to run.
- Does the documentation make sense?

Some things this PR does not do, but could in the future:
- Save the state of the plugins per-document. I.e. whenever you open a
  document the Eco tab starts with all plugins activated, regardless of how
  you used plugins the last time you opened the document.
- Expose many internals to the user.
- Allow you to check different plugins for different tabs.
- If a node has more than one _Footnote_ or _Tooltip_ annotation, the
  text from the annotations is concatenated. However, only one sort of
  annotation can make use of the heatmap overlay. To date, we have never had
  two plugins generating heatmap data, but if this were to happen the user
  would have to deactivate one of their plugins. This may be the right level
  flexibilty, but the alternative is to add some extra logic to Eco so that
  it chooses which data to display for the user.

Fixes #35


lib/eco/plugins/manager.py
```diff
@@ -0,0 +1,130 @@
...
+
+        # Number of plugins currently running.
+        self._plugins_running = 0
+
+    def _load_plugins(self):
+        """Load plugins and construct GUI elements.
+        """
+        # Load all plugins that can be found on disk.
+        name_space = dict(locals(), **globals())
+        for root, dirs, files in os.walk(self._plugin_dir):
```
ltratt, line 50: Do we want to recursively walk the dir? I'd have thought just the top-level (i.e. `os.listdir`) would be less surprising?

ltratt, line 50: Ah, I see a difference in expectation here. I was thinking that each plugin would be a directory.

ltratt:
OK, I broadly like the way this is going. My major initial comment is that I think plugins should be in their own directories (I can imagine some will have additional files as part of them, some of which might not even be Python code). Let's say we have a plugin `p`, then we could say something like `p/plug.py` is the plug-in's entry point?

ltratt:
Ah, I missed out that currently users have to copy plugins to `~/.eco/plugins`. That's a non-runner, as the default plugins we care about will get out of sync with main Eco. [I've no objection to `~/.eco/plugins` being searched in addition, but default plugins should be loaded from `lib/eco/plugins`.]

When I copy the directory, and try running things I get this problem:

```
UNIPYCATION=~/tmp/unipycation/pypy/goal/unipycation ~/src/eco/bin/eco
Couldn't import dot_parser, loading of dot files will not be possible.
  File "eco.py", line 1184, in <module>
    main()
  File "eco.py", line 1124, in main
    window=Window()
  File "eco.py", line 593, in __init__
    self.plugin_manager = PluginManager(self, self.ui.menuTool_plugins)
  File "/home/ltratt/src/eco/lib/eco/plugins/manager.py", line 28, in __init__
    self._load_plugins()
  File "/home/ltratt/src/eco/lib/eco/plugins/manager.py", line 52, in _load_plugins
    execfile(os.path.join(root, file_), name_space, name_space)
  File "/home/ltratt/.eco/plugins/__init__.pyc", line 1
SyntaxError: Non-ASCII character '\xf3' in file /home/ltratt/.eco/plugins/__init__.pyc on line 1, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details
```

snim2:
What is it you have copied to .eco there? I didn't touch any Unipycation
code...

ltratt:
I copied the plugins directory (the `UNIPYCATION` environment variable is a red herring) as per the instructions.

snim2:
There isn't (or shouldn't have been) a **init**.py in with the plugins. The
error message is very confusing, but the three plugin files should be
sl.py, cpython.py and jrubysl.py (from memory). If that is what you have in
.eco I think I will need to sit with your laptop to debug that one :(

ltratt:
This PR created a file `eco/plugins/__init__.py` (created by you as part of commit bf37502b8).

snim2:
I can't figure out how to get a url to a directory in a PR, but this is the
folder to copy:

https://github.com/snim2/eco/tree/eco-plugins/plugins

ltratt:
Oh, I see -- I didn't get that at all! Since we really don't want to make people copy anything to `~/.eco`, can't we make those things part of this PR?

snim2:
Are they really not in the PR? those files are only pushed to the pr branch?

ltratt:
Ah, they're at the top-level!

ltratt:
Ah, that dir is at the top level.

ltratt:
Though it then dies with

```
  File "eco.py", line 1184, in <module>
    main()
  File "eco.py", line 1124, in main
    window=Window()
  File "eco.py", line 593, in __init__
    self.plugin_manager = PluginManager(self, self.ui.menuTool_plugins)
  File "/home/ltratt/src/eco/lib/eco/plugins/manager.py", line 28, in __init__
    self._load_plugins()
  File "/home/ltratt/src/eco/lib/eco/plugins/manager.py", line 52, in _load_plugins
    execfile(os.path.join(root, file_), name_space, name_space)
  File "/home/ltratt/.eco/plugins/README.md", line 3
    A *plugin* in Eco is a proxy to an external tool, which can provide useful
                ^
SyntaxError: invalid syntax
```

because we `execfile` anything in that dir ;)

plugins/cpython.py
```diff
@@ -0,0 +1,98 @@
...
+    def __init__(self):
+        super(CPythonProfiler, self).__init__("Python 2.7.5",
+                                              "CPython profiler")
+
+    def run_tool(self):
+        f = tempfile.mkstemp()
+        self.export(f[1])
+
+        # python -m cProfile [-o output_file] [-s sort_order] myscript.py
+        proc = subprocess.Popen(["python2",
```
ltratt, line 41: I _think_ (but I might be wrong) that `python2.7` is a safer executable name (I have `python2.7` but not `python2` on my machine).

ltratt:
OK, I've got something running, sort of. I managed to enter a Python 2.7.5 program in and have the first line be profiled in cPython and displayed as a heatmap. However, I'm actually not quite sure how I managed to do it, and it hasn't worked since.

snim2:
The cpython profiler only reports on function calls btw

ltratt:
Somehow I got a heatmap out of a one line program `x = 2`. And now when I type in a function, I can't seem to get anything. Not quite sure why, to be honest!

snim2:
Do you have a **main**?

plugins/cpython.py
snim2, line 41: The language name needs to match the language in the "languages" list that Eco holds internally. At the moment the entry for Python is "Python 2.7.5", regardless of the Python 2 version installed on your machine. I agree that is a little strange, but it is documented in the README.

I pushed a fix to prevent the README from being exec'd this morning BTW,

ltratt, line 41: I mean the process name is `python2` but I think it should be `python2.7`.

ltratt:
No. Do I need a `__main__`?

snim2:
As it turns out this PR has a race hazard (between the code that annotates nodes and code in the main UI which can delete nodes). At first I thought this was fixable with a few more commits, but having given it a little more consideration I think this will need a re-write. So, I will close this PR now and open a new one in a few days, to avoid anyone wasting more time reviewing this code. I'll use some git magic to make sure code here is reused where possible and the comments are addressed, and I'll include a larger number of example plugins.

ltratt:
OK, thanks!

PR #58, opened 2015-11-30
FIX bug where stale heatmap data was not being removed

snim2:
Bug fix written up from work on Oracle video.


ptersilie:
How often is the stale data removed? If that happens really often there is maybe a way to remember all nodes that are annotated so we don't have to run over the whole file to find them later for removal. If it happens rarely though I'm not bothered.

snim2:
This PR removes the stale data just before new data is written onto the nodes. When I push a PR that includes the new plugin architecture I'll think again, but this PR corrects a mistake in the code which is currently in `master`.

Placing annotated nodes in a dictionary or similar is a good idea. I hadn't thought about it before, but I'll try it out for the upcoming plugin work.

ptersilie:
Fair enough. Merged. :)

PR #61, opened 2015-12-01
Export filename and linenumbers for PyHyp

### LINKED ISSUE #57: Implement support for extended PyHyp backtrace information.
The compilation functions used in PyHyp now accept additional arguments which allow the editor to feedback information which can later be used in the VM for backtraces. Although the VM could figure this stuff out itself at runtime, it is more convenient for the editor to supply this info as it already knows this ahead of time.

For the functions:
- `compile_py_func_global` 
- `compile_py_func`
- `compile_py_meth`
- `compile_php_func`

eco should pass two additional arguments:
- `filename` -- the filename in which the language box appears.
- `line_offset` -- the line number in `filename` in which the language box appears.

This is required for the PyHyp paper.

Many thanks

ptersilie:
Fixes #57

### DIFF
--- a/lib/eco/treemanager.py
+++ b/lib/eco/treemanager.py
@@ -1392,7 +1392,7 @@ class TreeManager(object):
         self.changed = True
         return
 
-    def fast_export(self, language_boxes, path):
+    def fast_export(self, language_boxes, path, source=None):
         # fix languagebox pointers
         for root, language, whitespaces in language_boxes:
             try:
@@ -1406,7 +1406,7 @@ class TreeManager(object):
         def x():
             return bos
         self.get_bos = x
-        return self.export(path)
+        return self.export(path, source=source)
 
     def load_file(self, language_boxes, reparse=True):
         # setup language boxes
@@ -1449,7 +1449,7 @@ class TreeManager(object):
             print("Grammar Error: could not determine grammar type")
             return
 
-    def export(self, path=None, run=False, profile=False):
+    def export(self, path=None, run=False, profile=False, source=None):
         for p, _, _, _ in self.parsers:
             if p.last_status == False:
                 print("Cannot export a syntacially incorrect grammar")
@@ -1464,7 +1464,7 @@ class TreeManager(object):
         elif lang == "HTML + Python + SQL":
             self.export_html_python_sql(path)
         elif lang == "PHP + Python" or lang == "PHP":
-            return self.export_php_python(path, run)
+            return self.export_php_python(path, run, source=source)
         elif lang == "Python 2.7.5":
             return CPythonExporter(self).export(path, run, profile)
         elif lang == "SimpleLanguage":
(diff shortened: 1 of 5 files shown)


vext01:
Hey Lukas,

I've just tested this branch. There's a few issues:
- 1) Language boxes are passing down a `None` filename.
- 2) Nested Language box filenames are off by one.
- 3) Top-level box line numbers are off by one, presumably due to the first `<?php` line that the user does not see.
- 4) All top-level filenames are the filename eco exported to.

1 is easy to fix. Not sure what to propose for the others. Any suggestions?

CC @ltratt 
eco

vext01:
Perhaps for 3 and 4, export to a filename matching (or similar) to that of the eco filename? E.g. `myfile.eco` becomes `myfile.php`. Maybe we can add an option to the VM to allow eco to pass a line number offset for outer PHP boxes. All a bit hacky -- will think some more.

vext01:
Another option to work around PHP line number offset is to require the user to input the enclosing `<?php` and `?>`. This would also allow interleaving of PHP with other text too. We must be careful however, to insert `{` and `}` inside each and every PHP block.

ptersilie:
I think the preview doesn't work properly. So try to export the file
properly. Also the line numbers are relative to the eco file which I
thought was what you wanted (so the user can see where the errors are when
he's writing in eco).

vext01:
Something odd is going on here. If I click execute in eco, it exports and runs this:

```
<?php{                                                                          
compile_py_func_global("def h():\n    raise ZeroDivisionError(\"my message\")", "None", 1);

function g() {                                                                  
    return h();                                                                 
}                                                                               

compile_py_func_global("def f():\n    return g()", "None", 8);                  

try {                                                                           
    f();                                                                        
} catch (PyException $e) {                                                      
    echo $e->getMessage() . "\n";                                               
    echo $e->getTraceAsString() . "\n";                                         
}                                                                               
}?>
```

If I use the menu to export:

```
<?php{
compile_py_func_global("def h():\n    raise ZeroDivisionError(\"my message\")", "/home/edd/research/softdev/pubs/fine_grain_lang_comp/eco_examples/exceptions.eco", 1);

function g() {
    return h();
}

compile_py_func_global("def f():\n    return g()", "/home/edd/research/softdev/pubs/fine_grain_lang_comp/eco_examples/exceptions.eco", 8);

try {
    f();
} catch (PyException $e) {
    echo $e->getMessage() . "\n";
    echo $e->getTraceAsString() . "\n";
}
}?>
```

The latter crashes the VM on my laptop, but it is quite an old build (will make a new one overnight).

Re, your comment above, the line numbers for inner boxes are indeed relative to the eco file (with an off-by-one), but note that we don't have the ability to name (or give line numbers) for the outer PHP box since they never hit a `compile_php_*` function and instead have the random unique filename that you temporarily export to. We would need to make sure that this filename matches the filename we pass to `compile_*` for inner boxes.

Do you follow?

It's not immediately obvious what we can do to fix this.

vext01:
Having re-built the VM, the version exported from the menu now runs, so that  is a red herring at-least:

```
wilfred:/tmp> ~/research/pyhyp/hippyvm.hippyvm/hippy-c /tmp/a.php 
ZeroDivisionError: my message
#0 /home/edd/research/softdev/pubs/fine_grain_lang_comp/eco_examples/exceptions.eco(3): h()
#1 /tmp/a.php(5): g()
#2 /home/edd/research/softdev/pubs/fine_grain_lang_comp/eco_examples/exceptions.eco(10): f()
#3 /tmp/a.php(11): <main>()
```

vext01:
exporter crashes:

```
  File "eco.py", line 1002, in exportAs
    self.getEditor().tm.export(path, source=ed.filename)
  File "/home/edd/research/eco/lib/eco/treemanager.py", line 1444, in export
    return self.export_php_python(path, run, source=source)
  File "/home/edd/research/eco/lib/eco/treemanager.py", line 1512, in export_php_python
    text = PHPPython.export(self.get_bos(), os.path.basename(source))
UnboundLocalError: local variable 'os' referenced before assignment
```

vext01:
Lukas, we should revisit this at the sprint.

vext01:
This seems to work on the paper example. Shall we merge? Do you want to squash?

lib/eco/treemanager.py
```diff
@@ -1518,28 +1518,35 @@ def export_html_python_sql(self, path):
...
+        import os
         if run:
             import tempfile
-            import os, sys, subprocess
-            f = tempfile.mkstemp()
-            os.write(f[0], PHPPython.export(self.get_bos()))
+            import sys, subprocess
+            d = tempfile.mkdtemp()
+            if source:
+                f = os.path.basename(source)
```
vext01, line 1528: Is 'f' closed?

vext01:
One late comment. Also some long lines in the diff, up to you if you re-work them.

vext01:
OK, I'm going to merge this due to time constraints, but please look into whether we are leaking file handles.

PR #63, opened 2015-12-10
Save a version for each change and store snapshots for undo/redo

ptersilie:
From now on each change in Eco is saved as it's own version. Several changes are grouped into snapshots, which can be loaded through undo/redo.

### DIFF
--- a/lib/eco/treemanager.py
+++ b/lib/eco/treemanager.py
@@ -291,6 +291,7 @@ class TreeManager(object):
         self.savenextparse = False
         self.saved_lines = {}
         self.saved_parsers = {}
+        self.undo_snapshots = []
 
         self.tool_data_is_dirty = False
 
@@ -628,7 +629,14 @@ class TreeManager(object):
 
     def key_shift_ctrl_z(self):
         self.log_input("key_shift_ctrl_z")
-        if self.get_max_version() > self.version:
+        try:
+            i = self.undo_snapshots.index(self.version)
+            if i == len(self.undo_snapshots) - 1:
+                return
+            undo_amount = self.undo_snapshots[i+1] - self.undo_snapshots[i]
+        except ValueError:
+            undo_amount = self.undo_snapshots[0] - self.version
+        for i in range(undo_amount):
             self.version += 1
             TreeManager.version = self.version
             self.recover_version("redo")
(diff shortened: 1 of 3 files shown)


lib/eco/treemanager.py
```diff
@@ -628,7 +629,14 @@ def jump_to_error(self, parser):
 
     def key_shift_ctrl_z(self):
         self.log_input("key_shift_ctrl_z")
-        if self.get_max_version() > self.version:
+        try:
+            i = self.undo_snapshots.index(self.version)
+            if i == len(self.undo_snapshots) - 1:
+                return
+            undo_amount = self.undo_snapshots[i+1] - self.undo_snapshots[i]
+        except ValueError:
```
ltratt, line 637: Is catching a `ValueError` the best way of doing this? I can't even work out what would throw such an exception. Is it `undo_snapshots`? Could we put some sort of check (e.g. `if len(X) > 0:`) instead of the `try`?

ltratt:
I _think_ `undo_snapshots` can throw a ValueError, but I'm really unclear as to why. Otherwise, looks OK.

lib/eco/treemanager.py
ptersilie, line 637: It's `index` that throws the `ValueError`. Not sure there is an alternative for checking if a value is not being in `undo_snapshots`. I guess I could do

```
if self.version in self.undo_snapshots:
    i = self.undo_snapshots.index(self.version)
```

but I'm not sure that's any better as it searches `undo_snapshots` twice.

ltratt, line 637: Ah, so it's a performance hack? How big does `undo_snapshots` tend to get?

ptersilie, line 637: Not massively huge. As big as how many undo operations you have in your document. The longer you type, the bigger. You are more likely to run out of memory than making running over `undo_snapshots` slow. It's not a hack. I just think it's bad programming to iterate over a list twice when it can be done in one.

ltratt, line 637: It's an unfortunate part of Python's API. Such is life.

ltratt:
OK, so ready to merge?

lib/eco/treemanager.py
ptersilie, line 637: I could maybe move parts in the finally block to make clear what throws the error, e.g.:

```
        try:
            i = self.undo_snapshots.index(self.version)
        except ValueError:
            ...
        finally:
            if i == len(self.undo_snapshots) - 1:
                return
            undo_amount = self.undo_snapshots[i+1] - self.undo_snapshots[i]
```

ltratt, line 637: I don't think that's semantically equivalent, because currently you don't assign to `i` in the `except`?

ptersilie, line 637: The except returns, so it doesn't matter. But I'm happy to just leave it as it is.

I think it's ready to merge.

PR #65, opened 2016-01-04
Allow setting PYHYP/UNIPYCATION paths from within Eco

ptersilie:
Added a from to the settings that allows the user to specify the paths to pyhyp, pypy_prefix and unipycation within Eco.

### DIFF
--- a/lib/eco/eco.py
+++ b/lib/eco/eco.py
@@ -293,6 +293,10 @@ class SettingsView(QtGui.QMainWindow):
         self.connect(self.ui.heatmap_low, SIGNAL("clicked()"), self.pick_color)
         self.connect(self.ui.heatmap_high, SIGNAL("clicked()"), self.pick_color)
 
+        self.connect(self.ui.btpyhyp, SIGNAL("clicked()"), self.choose_file)
+        self.connect(self.ui.btpypyprefix, SIGNAL("clicked()"), self.choose_dir)
+        self.connect(self.ui.btunipycation, SIGNAL("clicked()"), self.choose_file)
+
         self.foreground = None
         self.background = None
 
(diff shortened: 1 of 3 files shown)


ltratt:
What does `pypy_prefix` apply to?

ptersilie:
It's the path to the pypy checkout switched to the hippy_bridge branch. PyHyp needs it to run.

ltratt:
Ah, I see (although weirdly I don't need to set that... at least on one of my older builds of PyHyp). Maybe we can make this a bit clearer and instead of having a section "environment variables" (which is a bit confusing, given that they are no longer environment variables), we can have sections "PyHyp" (with settings "PyHyp binary" and "PyPy libraries" (or similar)) and "Unipycation" ("Unipycation binary") or similar?

ptersilie:
I called it environment variables, because I basically add the provided paths to `os.env`. Also if the environment variables are set within the OS already, one doesn't have to provide them inside of Eco. Also the error message when trying to execute stuff from within Eco is "Environment variable PYHYP not set".

ltratt:
But Eco is the thing that reads those environment variables :) So we're setting `os.env` in one part of Eco only to read it in another part of Eco (`lib/eco/treemanager.py`). I think we can probably get rid of all of these environment variables and make them normal attributes within Eco?

ptersilie:
Sure we can do that. I just did it this way so people who have already set the environment variables in their OS don't need to set them again within Eco.

ltratt:
I think it's probably best to remove the environment stuff, otherwise people might define the variable outside Eco, then define it inside, and be confused as to which actually effects the outcome.

ptersilie:
Ok

ltratt:
Thanks!

ptersilie:
@ltratt Can you have another look?

ptersilie:
Where? In the PR title?

ltratt:
I verified this works with PyHyp, but not Unipycation. This is because the connect4 example I have doesn't parse against the current grammar. Fixing it caused weird redraw errors which I haven't investigated further, but that's a problem for another time.

So I think this PR is definitely heading in the right direction. We still need to fix this so that the settings don't say "environment variables" and things are split up into "PyHyp" and "Unipycation" settings (I really had to scratch my head to remember what PyPy-prefix is...).

ltratt:
Looks good to me! Do you want to squash anything?

ptersilie:
Yes, will squash now.

ltratt:
OK. Let me know when you're done.

ptersilie:
Done.

ltratt:
Good stuff!

PR #66, opened 2016-01-12
Incremental Parser extension plugins

ptersilie:
WORK IN PROGRESS

I started thinking about and working on adding plugins to extend the incremental parser instead of adding them directly into the parser. I removed the indentation extension to support Python and put it in a separate file which is then loaded by the incremental parser.

I'm not sure however about the plugin API. This PR is a first attempt and I'd be grateful for some suggestions on how to make it more intuitive and flexible.

### DIFF
--- a/lib/eco/incparser/incparser.py
+++ b/lib/eco/incparser/incparser.py
@@ -33,6 +33,7 @@ from syntaxtable import SyntaxTable, FinishSymbol, Reduce, Goto, Accept, Shift
 from stategraph import StateGraph
 from constants import LR0, LR1, LALR
 from astree import AST, TextNode, BOS, EOS
+from ip_plugins.plugin import PluginManager
 
 import logging
 
@@ -97,11 +98,11 @@ class IncParser(object):
         self.anycount = set()
         self.status_by_version = {}
         self.errornode_by_version = {}
-
-        self.comment_tokens = []
-
-        self.indent_stack = None
         self.indentation_based = False
+
+        self.pm = PluginManager()
+        self.pm.loadplugins(self)
+        self.pm.do_incparse_init()
 
         self.previous_version = None
         logging.debug("Incemental parser done")
@@ -125,12 +126,7 @@ class IncParser(object):
                 pickle.dump(self.syntaxtable, open(filename, "w"))
 
         self.whitespaces = whitespaces
-        if not rules:
-            print("Warning: incparser has not access to comment tokens")
-        elif rules.has_key(Nonterminal("comment")):
-            rule = rules[Nonterminal("comment")]
-            for a in rule.alternatives:
-                self.comment_tokens.append(a[0].name)
+        self.pm.do_incparse_from_dict(rules)
 
     def init_ast(self, magic_parent=None):
         bos = BOS(Terminal(""), 0, [])
@@ -156,33 +152,14 @@ class IncParser(object):
         self.undo = []
         self.current_state = 0
         self.stack.append(Node(FinishSymbol(), 0, []))
-        self.stack[0].indent = [0]
         bos = self.previous_version.parent.children[0]
         self.loopcount = 0
         self.anycount = set()
         self.any_newlines = []
-        self.last_indent = [0]
 
         USE_OPT = True
 
-        eos = self.previous_version.parent.children[-1]
-        d = eos.prev_term
-        while isinstance(d.symbol, IndentationTerminal):
-            d = d.prev_term
-        self.last_token_before_eos = d
-        if isinstance(d, BOS):
-            # if file is empty, delete left over indentation tokens
-            n = d.next_term
-            while isinstance(n.symbol, IndentationTerminal):
-                n.parent.remove_child(n)
-                n = n.next_term
-
-        # fix indentation after bos. Should result in an error for whitespace
-        # at the beginning
-        if bos.next_term.lookup == "<ws>":
-            bos.insert_after(TextNode(IndentationTerminal("INDENT")))
-        elif isinstance(bos.next_term.symbol, IndentationTerminal):
-            bos.next_term.parent.remove_child(bos.next_term)
+        self.pm.do_incparse_inc_parse_top()
 
         la = self.pop_lookahead(bos)
         while(True):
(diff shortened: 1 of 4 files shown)


ltratt:
I might be an idiot, but I can see lots of code deleted, but not much code added. Is there a prototype plugin that's not been added to the PR?

ptersilie:
Ach damnit. Of course I forgot to add the new files. :/

lib/eco/plugins/pythonindentation.py
```diff
@@ -0,0 +1,305 @@
...
+    def set_total_indent(self, node):
+        l = []
+        if node.children:
+            for c in node.children:
+                if c.indent:
+                    l = c.indent
+        if l:
+            node.indent = l
+
+def load(caller):
```
ltratt: Is this function ever called?

lib/eco/plugins/pythonindentation.py
```diff
@@ -0,0 +1,305 @@
...
+
+    def incparse_end_any(self, newlines):
+        # update succeeding
+        if self.incparser.indentation_based:
+            for n in newlines:
+                self.update_succeeding_lines(n, self.last_indent[-1], list(self.last_indent))
+
+    def incparse_from_dict(self, rules):
+        if not rules:
+            print("Warning: incparser has not access to comment tokens")
```
ltratt: s/not/no/

ltratt:
This looks like a very good start to me, and looks very much like what I was thinking of. My guess is that the only way to know for sure what the "perfect" API might be is to write another plugin. I'm sure the API will evolve a bit, but my guess is that what you've started with here will not change _that_ much.

One question that I don't have a good answer to: how do plugins interact? My guess is that some plugins are mutually exclusive, and running them together would cause weird stuff to happen. Maybe we could even be restrictive and say there's a maximum of one active plugin at a time?

lib/eco/plugins/pythonindentation.py
ptersilie: Yes, this is called by pluginmanager.loadplugins (https://github.com/softdevteam/eco/commit/e5bdc1a3e885cfbbf7486c360ab9d65cbe9ee8c1#diff-581d80aa746ab35aa19b6f5e7c05c274R13)

ptersilie: So plugins load themselves in a sense which also gives the plugin author an opportunity to add constraints on when to actually load the plugin or not

ptersilie:
At the moment plugins can't interact. And I'm not sure how this would be done. It certainly would require the current API to change a lot. At this point I'm not sure if this is even needed/wanted.

ptersilie:
I guess you mean one plugin per language? Yes that might be an option (at least for now).

ltratt:
Yes, one plugin per language.

ptersilie:
Yes, maybe that's a good restriction for the first implementation.

ltratt:
OK, then, with that restriction added, we might have something that's almost ready for merging? Or do you want to do another plugin? Or ...?

snim2:
@ptersilie you and I are both creating "plugins" and plugin managers for different things. I'll be raising a new PR for my stuff in a few days, but we have both put our plugin manager stuff into `lib/eco/plugins/`, so I expect we need to change that.

Since you have pushed first, it makes sense for me to move my stuff to a new package. Maybe `lib/eco/visualisations/` or some such?

ltratt:
I'd suggest Lukas calls his stuff something like "ip_plugins", since it's totally restricted to the incremental parser.

ptersilie:
Sure, I can put this into another folder. It looks like we are going to have different sorts of plugins. Although I coded my plugin api to allow to be used from anywhere within eco, which means they don't have to be restricted to the incremental parser. But for now I'm happy with that restriction.

snim2:
Hmm. Where are you expecting users to store plugins?
Perhaps we should wait until both PRs are merged, then try to unify the API. I think it will be difficult to do that as we go along...

ltratt:
I don't think it's worth worrying too much about having "one plugin" type, at least not yet. The sorts of things an incremental parser plugin does are quite different from the visualiser plugins.

ptersilie:
@snim2 Yes, let's think about merging the APIs once both are merged
@ltratt Yes they are different but since the plugins are just function calls (at least for my api) it would be absolutely no problem to merge both plugin APIs. The plugins just need to activate themselves appropriately and implement the correct methods.

ptersilie:
Okay, moved everything to a new folder. If you are happy I'll squash and you can merge this. I reckon this will be subject to more change once we have more plugins. For now it's difficult to reason about how to design this API as we only have one plugin. Once we add another one we can update the API.

ltratt:
Squash away and we can merge.

ptersilie:
Oh, what about the restriction though? I haven't done that now as I'm unsure how, e.g. where should a user define which plugin their language has to load, etc.

My suggestion would be to leave it for now and solve it once we have two plugins. Maybe by then we changed out mind and don't like the one-plugin restriction anymore.

snim2:
@ptersilie I don't know if this is helpful to you, but I am storing the visualisation plugins in `~/.eco/`. Once I have tidied everything up, my plan was to add an entry in settings to allow the user to choose a different directory.

ltratt:
@ptersilie, doing the restriction later is fine with me.

ltratt:
@snim2, we said we wouldn't store (at least the default) plugins in `~/.eco`, as that means they're not properly versioned and synced with the main editor.

snim2:
@ltratt well, the way I had it the "default" plugins were in the repo and the user had to copy them to their own directory. For most of these plugins, the user will want to change some paths and so on, so the "default" plugins cannot be guaranteed to work for anyone. The exception to this is the CPython profiler which really does only use the Python stdlib.

ltratt:
We really don't want users copying code -- down that way lies disaster. If there are things they need to configure, we can have that in some sort of config file (maybe even, as a quick hack, a Python file which Eco executes), but the default plugins really must remain in the repo.

ptersilie:
Yes, the indentation plugin is pretty mandatory for eco in order to run Python so I'd like to have this within the repo.

Why don't we just have two plugin directories, e.g. always check in lib/eco/plugins (standard plugins) and then additionally check in .eco/plugins (user plugins). The user can then decide if they want to store other plugins within the user or the main folder.

ltratt:
Yes, I think it's fine to search in an external dir for plugins the user may have written. I just think we mustn't make them copy code out of the Eco release into that external directory.

snim2:
@ltratt that is a somewhat different model for what a plugin is.

ltratt:
@snim2 There is more than one common model of plugins ;) Anyway, this is something for the other PR -- we should probably move this conversation to your upcoming PR.

ptersilie:
squashed.

ltratt:
@ptersilie Ready to merge?

ptersilie:
Yep

ptersilie:
Maybe wait for the tests

ltratt:
Good point.

PR #67, opened 2016-01-13
Added LSpace viewer that spawns an L-Space viewer as a separate process.

Britefury:
This PR adds a 'View in LSpace' item to the View menu.

It spawns the LSpace 'eco_viewer' example program to display the content of the currently visible document.

### DIFF
--- a/lib/eco/lspace_ext/lspace.py
+++ b/lib/eco/lspace_ext/lspace.py
@@ -1,0 +1,107 @@
+import os, json, subprocess
+
+
+class CannotFindLSpaceRootError (Exception):
+    pass
+
+class CannotLaunchLSpaceError (Exception):
+    pass
+
+
+def _valid_lspace_root(lspace_root=None):
+    if not os.path.exists(lspace_root):
+        raise CannotFindLSpaceRootError("Cannot find LSpace directory; you "
+                "can set the LSpace root using the settings dialog")
+    return lspace_root
+
+def _get_lspace_viewer_path(lspace_root=None):
+    lspace_root = _valid_lspace_root(lspace_root)
+    return os.path.join(lspace_root, 'target', 'release', 'examples', 'json_viewer')
+
+
+
+class JSONData (object):
+    def __init__(self, **js):
+        self.js = js
+
+    def as_json(self):
+        return self.js
+
+
+class FlowIndent (JSONData):
+    @staticmethod
+    def no_indent():
+        return FlowIndent(indent_type='no_indent')
+
+    @staticmethod
+    def first(indent):
+        return FlowIndent(indent_type='first', indent=indent)
+
+    @staticmethod
+    def except_first(indent):
+        return FlowIndent(indent_type='except_first', indent=indent)
+
+
+
+class Pres (JSONData):
+    @staticmethod
+    def coerce(x):
+        if isinstance(x, Pres):
+            return x
+        else:
+            present = getattr(x, '__present__', None)
+            if present is None:
+                return Text(repr(x))
+            return present()
+
+
+
+class Text (Pres):
+    def __init__(self, text):
+        super(Text, self).__init__(__type__='Text', text=text)
+
+
+class Column (Pres):
+    def __init__(self, children, y_spacing=0.0):
+        super(Column, self).__init__(__type__='Column', y_spacing=y_spacing,
+                children=[Pres.coerce(c).as_json() for c in children])
+
+
+class Row (Pres):
+    def __init__(self, children, x_spacing=0.0):
+        super(Row, self).__init__(__type__='Row', x_spacing=x_spacing,
+                children=[Pres.coerce(c).as_json() for c in children])
+
+
+class Flow (Pres):
+    def __init__(self, children, x_spacing=0.0, y_spacing=0.0, indentation=None):
+        if indentation is None:
+            indentation = FlowIndent.no_indent()
+        super(Flow, self).__init__(__type__='Flow', x_spacing=x_spacing,
+                y_spacing=y_spacing, indentation=indentation.as_json(),
+                children=[Pres.coerce(c).as_json() for c in children])
+
+
+def viewer(pres, lspace_root=None):
+    """
+    Start the LSpace eco_viewer external tool
+
+    pres: a `Pres` instance to show or another object that will be coerced
+    into a `Pres`.
+
+    lspace_root: [optional] the root directory of the LSpace repo
+    """
+    lspace_viewer_path = _get_lspace_viewer_path(lspace_root)
+    lspace_viewer_dir = os.path.split(lspace_viewer_path)[0]
+
+    try:
+        proc = subprocess.Popen([lspace_viewer_path], cwd=lspace_viewer_dir,
+                                stdin=subprocess.PIPE)
+    except OSError:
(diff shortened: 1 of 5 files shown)


snim2:
`RuntimeError: Could not execute LSpace viewer at /home/snim2/kcl/lspace/target/release/examples/eco_viewer, set its path using the LSPACE_VIEWER_PATH environment variable`

Maybe worth adding the `LSPACE_VIEWER_PATH` to the new settings pane that @ptersilie recently committed?

Britefury:
Okay, I'll go do that :)

snim2:
Also setting that env var to be the root of a clone of the lspace repo does not work, which surprised me a little!

Britefury:
Good point, thanks! :)

Britefury:
Done; now uses the settings and the path provided is the repo root rather than the executable root as per @snim2's suggestion.

Britefury:
I have modified the LSpace viewer to account for changes to the Rust side.

Prior to these changes, Eco described what it wanted in terms of a document being composed of lines, which were composed of tokens, which were textual items; similar in some ways to the structure of an Eco document.

The new system allows a generic presentation to be described. It is serialised to JSON and sent to the Rust app for rendering. As a result, we can now display whatever we like. The presentation API is a taster of the API that we expect to develop in the future.

snim2:
OK, currently blocked on Issue 29 in the LSpace repo: https://github.com/softdevteam/lspace/issues/29

lib/eco/lspace_ext/lspace.py
```diff
@@ -0,0 +1,126 @@
...
+        super(Flow, self).__init__(__type__='Flow', x_spacing=x_spacing, y_spacing=y_spacing, indentation=indentation.as_json(),
+                                   children=[Pres.coerce(c).as_json() for c in children])
+
+
+def viewer(pres, lspace_root=None):
+    """
+    Start the LSpace eco_viewer external tool
+
+    :param pres: a `Pres` instance to show or another object that will be coerced into a `Pres`.
+    """
```
snim2, line 93: I think `:param` is syntax for one of the Python Javadoc equivalents. I don't _think_ we are using this elsewhere in Eco, so probably best to rewrite this in prose.

lib/eco/lspace_ext/view_ecodoc_in_lspace.py
```diff
@@ -0,0 +1,91 @@
...
+        return Column(self.lines)
+
+
+
+def view_in_lspace(tree_manager, lspace_root=None):
+    """
+    View the content of the given `TreeManager` in the LSpace eco_viewer tool
+    :param tree_manager: a `TreeManager` instance
+    :return:
+    """
```
snim2, line 67: likewise.

lib/eco/lspace_ext/lspace.py
```diff
@@ -0,0 +1,126 @@
...
+                                    cwd=lspace_viewer_dir, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        except OSError:
+            raise CannotLaunchLSpaceError("Could not execute LSpace viewer at "
+                "{0}, you can set the LSpace root using either the settings "
+                "dialog or the LSPACE_ROOT environment variable".format(
+                lspace_viewer_path))
+
+        out, err = proc.communicate()
+        print out
+        if len(err) > 0:
```
snim2: Why print to STDOUT?

lib/eco/lspace_ext/lspace.py
```diff
@@ -0,0 +1,126 @@
...
+    a_fd, a_path = tempfile.mkstemp(suffix='.json')
+
+    try:
+        a_f = os.fdopen(a_fd, 'w')
+        json.dump(Pres.coerce(pres).as_json(), a_f)
+        a_f.close()
+
+        try:
+            proc = subprocess.Popen([lspace_viewer_path, a_path],
+                                    cwd=lspace_viewer_dir, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
snim2: OK, so this is where you actually launch the external viewer. I guess the GUI window is actually provided by the Rust code. At present, you are running the LSpace viewer modally, so the viewer blocks the main GUI thread in Eco, and in Ubuntu the main window is greyed out a little (because it cannot accept events, etc). I'm not sure whether this is a good thing or not. If you take the view that an end user should not be able to edit their document or open another Eco tool while looking at the LSpace viewer, then this code is fine. If you think the user should be able to keep the LSpace viewer open and still use Eco, then you need to change this code a bit. The LSpace viewer would need to run in its own thread or process (e.g. with http://pyqt.sourceforge.net/Docs/PyQt4/qthread.html).

lib/eco/lspace_ext/lspace.py
Britefury, line 93: Fixed.

lib/eco/lspace_ext/view_ecodoc_in_lspace.py
Britefury, line 67: Fixed.

lib/eco/lspace_ext/lspace.py
```diff
@@ -0,0 +1,126 @@
...
+        except OSError:
+            raise CannotLaunchLSpaceError("Could not execute LSpace viewer at "
+                "{0}, you can set the LSpace root using either the settings "
+                "dialog or the LSPACE_ROOT environment variable".format(
+                lspace_viewer_path))
+
```
snim2: this formatting looks a bit odd to me. Might be worth looking at how string substitution is formatted elsewhere in Eco (I think perhaps we don't use `format()`?)

lib/eco/lspace_ext/lspace.py
Britefury: Okay, an exception is raised if errors are detected. Otherwise the output is printed to STDOUT as it can show one or two interesting things.

snim2: OK, but this is something that you will need but the user will not, I think? So, I would suggest not committing it. Alternatively, we could add a global `--debug` flat on the command line, for things like this.
/cc @ptersilie

lib/eco/lspace_ext/lspace.py
```diff
@@ -0,0 +1,126 @@
...
+
+if 'LSPACE_ROOT' in os.environ:
+    LSPACE_ROOT = os.environ['LSPACE_ROOT']
+else:
+    LSPACE_ROOT = os.path.expanduser(os.path.join(
+        '~', 'kcl', 'lspace'))
+
+
+def _valid_lspace_root(lspace_root=None):
```
ptersilie, line 11: I think @ltratt  and I agreed to only have one path pointing to external stuff to no confuse the user with multiple settings. In Eco's case we decided to only use the settings and not use environment vars at all.

Quote from https://github.com/softdevteam/eco/pull/65:

> I think it's probably best to remove the environment stuff, otherwise people might define the variable outside Eco, then define it inside, and be confused as to which actually effects the outcome.

ptersilie:
Okay looks good to me. Maybe some squashing and then I'm ready to merge this.

Britefury:
Okay, will do.

Britefury:
Squashed.

snim2:
@ptersilie do you want to assign yourself to this? I don't mind either way

ptersilie:
@snim2 Ah, didn't know you were assigned to this.

PR #68, opened 2016-01-19
Added initial version of a JavaScript grammar

ptersilie:
UNDER DEVELOPMENT


ptersilie:
@jaheba Once the tests finish this would be ready to merge.

jaheba:
``` javascript
var a = function b(){};
----
function f() {
  //foo
}
----
function f() {
  "use strict";
}
----
var b = {"b": 2};
----
var c = {c: "3"};
----
function f(a, b){
    return;
}
----
var $x;
----
var x = {foo:bar,};
----
var x = {1:2,};
```

This is all valid Javascript but is marked invalid by the parser.

ptersilie:
I fixed a few of these like

``` javascript
var a = function b(){};
```

and all string and comment related ones. The others I currently don't know how to fix. So for the ones with the commas, just remove the comma for now. And the `var $x` one I'm not even sure what JavaScript version that is, so it might not be allowed in the version our grammar is from.

jaheba:
What do you mean with JavaScript version? JavaScript itself is no specification, Ecmascript is. And `$` has long been a part of it or else we wouldn't have jquery which provides most of its API through `$`. In contrast to other languages `$` is perfectly allowed in the same sense as `_` is.

jaheba:
And shouldn't this fix the comma thing?

```
objectLiteral ::=
      "LBRACE" "RBRACE"
    | "LBRACE" propertyNameAndValueListWithOptionalComma "RBRACE"
    ;

propertyNameAndValueListWithOptionalComma ::=
   propertyNameAndValueList
 | propertyNameAndValueList "COMMA"
;
```

ptersilie:
Yes, I meant EcmaScript version. I didn't know that `$` is just part of the identifier name. In that case this is an easy lexer fix. I just couldn't find `$` in any of the grammars so I wasn't sure what it meant.

jaheba:
Sorry the example might look a bit confusing with a slight resemblance to PHP.

ptersilie:
Good point about the comma problem. The solution was actually even simplier:

```
objectLiteral ::=
      "LBRACE" "RBRACE"
    | "LBRACE" propertyNameAndValueList "RBRACE"
    | "LBRACE" propertyNameAndValueList "COMMA" "RBRACE"
    ;
```

jaheba:
Looks good so far except for `var b = {"b": 2};`.

In Javascript keys in objects are always strings. The syntax even allows that identifiers which are normally considered a keyword are fine in the object definition. For example `x = {for: 0}` is valid since `for` is considered as the string `"for"`. Since keys are strings it is also allowed to do stuff like `x = {"Hello World": 42}` which wouldn't be possible without the quotation marks.

Thus the key can be either written as a valid identifier (letters, numbers, underscores and dollar signs) without accounting for keywords or using a string literal.

Edit: On a second note this looks more or less fine, however it doesn't work:

```
propertyName ::=
      "identifier"
    | "string"
    | "decimal"
    ;
```

Edit 2: Alright, there is no rule called `"string"`, only `"single_string"`.

Edit 3:
I think this is the right solution:

```
propertyName ::=
    "identifier"
    | literal
```

jaheba:
`identifier: "[a-zA-Z_$][a-zA-Z0-9_]*"` should be `identifier: "[a-zA-Z_$][a-zA-Z0-9_$]*"`

jaheba:
`"hex"` is not defined, and thus `0x1` is not allowed.

ptersilie:
I think it should rather be

```
propertyName ::=
    "identifier"
    | literal
    | single_string
```

or otherwise we can't have strings as propertynames, right?

ptersilie:
So `$$$$$` is a valid identifier in JavaScript!?

ptersilie:
Ah, yes I haven't added hex values. You are welcome to do this if you want. :)

jaheba:
You bet.

jaheba:
> I think it should rather be [ ... ] or otherwise we can't have strings as propertynames, right?

Literals include strings:

```
literal ::= "decimal" | single_string | "null" | "bool" | "hex";
```

jaheba:
Regarding the hex thing:

Now we are at the true joys of Javascript vs EcmaScript. ES 5 didn't have literals for hex values, ES 6 does. And since valid Javascript is more or less defined as what is currently running, Javascript now also has them too, so I would propose to add them along side with binaries (`0b`) and octals (`0o`).

jaheba:
- [ ] Division: `1/2`
- [ ] "#hello"
- [ ] `1.2`;
- [ ] `var _ = 3`;
- [ ] `var $ = 4`;
- [ ] `var $_$ = 5`;
- [ ] `var x = 6 > 7`;

ptersilie:
@jaheba I suggest we merge this branch first. Then you can create a new branch/PR with all the lexing fixes for Python and JavaScript. Sounds like a plan?

jaheba:
Ok.

PR #70, opened 2016-01-21
Editor optimisations

ptersilie:
Some optimisations to make the drawing of source code faster.


ltratt:
Looks good to me. Do you want to squash then I can merge?

ptersilie:
I'd like to have these commits separate as they are not really related. Should have been two different PRs really.

Let's wait for tests to run in any case.

PR #72, opened 2016-02-04
Only skip drawing space as prolog's comments are whitespace

### LINKED ISSUE #71: Prolog comments cause remainder of line to be invisible
In a prolog box, if you type:

```
p(X, Y) % this is a call to p
```

Then everything after and including the percent is invisible.

Spoke with Lukas, apparently the % is being considered a whitespace character, and these are not drawn to improve performance.

Cheers

ptersilie:
Fixes #71

### DIFF
--- a/lib/eco/editor.py
+++ b/lib/eco/editor.py
@@ -88,7 +88,7 @@ class NormalEditor(Editor):
             paint.setPen(QPen(QColor(highlighter.get_color(node))))
             self.setStyle(paint, highlighter.get_style(node))
             text = node.symbol.name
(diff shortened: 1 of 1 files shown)


vext01:
Works for me.

One question: how many symbols begin with `" "`? Could the whole check be:

```
 if node.symbol.name != " ":
   actually_draw()
```

Cheers

ptersilie:
The problem is that whitespace tokens can be any amount of " " (space).

vext01:
I see. Then I'm OK with this.

PR #73, opened 2016-02-09
Supporting comments/strings through grammar rewriting

ptersilie:
Replaced the previous "AnySymbol"-approach with a new approach that rewrites the grammars to allow the use of multiline structures such as strings and comments while keeping changes to the parser minimal.


lib/eco/grammar_parser/bootstrap.py
```diff
@@ -149,21 +190,45 @@ def parse_precedence_symbols(self, symbol):
...
+                    r.add_alternative([])
+                    self.rules[r.symbol] = r
+                # remove whitespace before special rule from parent rule, e.g.
+                # multistring ::= "MLS" WS *match_until "MLS" WS
+                #                       ^ this WS causes shift/reduce conflicts
+                prule = self.rules[Nonterminal(parentrule)]
+                for a in prule.alternatives:
+                    for i in range(len(a)):
+                        sym = a[i]
+                        if sym.name == "WS":
```
ltratt, line 213: Can this not be `if len(a) > i + 1` and remove the `try ... except`?

lib/eco/grammar_parser/bootstrap.py
```diff
@@ -256,14 +319,34 @@ def parse_symbol(self, node):
...
+        s = []
+        for c in symbol.children:
+            if c.symbol.name == ",":
+                continue
+            if c.symbol.name == "WS":
+                continue
+            if c.symbol.name == "f_args":
+                rec_s = self.parse_fargs(symbol.children[0])
+                s.extend(rec_s)
+            if c.lookup == "terminal":
```
ltratt: Can the `c.symbol.name` and `c.lookup` clauses both be true?

ltratt:
Only a few minor comments from me...

ptersilie:
@ltratt i think this would be ready to merge

ltratt:
Did you rebase? That makes it really difficult to compare what's changed.

ptersilie:
Yes, I rebased. But you already checked this repo. I just applied your comments and then rebased.

ltratt:
So I have one, potentially minor, comment (see above).

ptersilie:
You are right. Fixed.

ltratt:
OK, so let's squash, then I can merge.

ptersilie:
Squashed.

ltratt:
Thanks!

Issue #80, opened 2016-03-09
shift deletes current selection


jaheba:
When `shift` on its own is pressed, it deselects the current selection.

PR #81, opened 2016-03-09
Improved arrow key handling on selected text.

### LINKED ISSUE #78: arrow keys behaviour on selected text
When some text is selected the left and right keys should jump to the start or end of the selection. Currently it jumps to the next/previous character instead.

Pressing up should jump a line upwards, but taking the start of the selection as reference.

jaheba:
Fixes #78

TODO:
- [x] on key left, jump to left start of selection, not where selection has started (right to left selection) [same for right]
- [x] The line position is not considered, only the cursor column.
- [x] Same with crossing language boxes.

### DIFF
--- a/lib/eco/nodeeditor.py
+++ b/lib/eco/nodeeditor.py
@@ -723,8 +723,6 @@ class NodeEditor(QFrame):
         self.show_cursor = True
 
         if e.key() in [Qt.Key_Shift, Qt.Key_Alt, Qt.Key_Control, Qt.Key_Meta, Qt.Key_AltGr]:
-            if e.key() == Qt.Key_Shift:
-                self.tm.key_shift()
             return
 
         text = e.text()
--- a/lib/eco/treemanager.py
+++ b/lib/eco/treemanager.py
@@ -1018,7 +1018,6 @@ class TreeManager(object):
                 self.selection_start = self.cursor.copy()
             self.cursor_movement(key)
             self.selection_end = self.cursor.copy()
-
         else:
             if self.hasSelection():
                 self.jump_cursor_within_selection(key)
--- a/lib/eco/treemanager.py
+++ b/lib/eco/treemanager.py
@@ -996,8 +996,6 @@ class TreeManager(object):
 
     def key_shift(self):
         self.log_input("key_shift")
-        self.selection_start = self.cursor.copy()
-        self.selection_end = self.cursor.copy()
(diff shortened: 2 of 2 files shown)


ptersilie:
There is one little remaining bug: When you select text from right to left (instead of left to right) and then use the cursor keys, the cursor jumps to the wrong end.

jaheba:
Two more issues:
- [ ] The line position is not considered, only the cursor column.
- [ ] Same with crossing language boxes.

The _real_ position of the cursor has to considered.

jaheba:
@ptersilie Can you have a look?

jaheba:
Don't merge yet, let's do #80 in this too.

lib/eco/treemanager.py
```diff
@@ -996,8 +996,6 @@ def key_delete(self):
 
     def key_shift(self):
```
ptersilie, line 997: Since `key_shift` is not used anymore I think we can delete it completely.

lib/eco/nodeeditor.py
```diff
@@ -729,8 +729,6 @@ def keyPressEvent(self, e):
         self.show_cursor = True
 
         if e.key() in [Qt.Key_Shift, Qt.Key_Alt, Qt.Key_Control, Qt.Key_Meta, Qt.Key_AltGr]:
```
ptersilie, line 725: Since `key_shift` is dead, it looks like this whole block can go, right?

ptersilie:
Added some comments about unused code we need to get rid of.

lib/eco/nodeeditor.py
jaheba, line 725: I handle this in #88

ltratt:
One minor thing I noticed: most of eco uses the double quote `"` for strings, whereas this PR uses single quotes `'`. Personally, I have no strong opinion about which we use, but we should be consistent within a single project.

ptersilie:
Should we? I don't really care if it's all odds and sodds.

ltratt:
Yes, we should :) Code consistency is a virtue (when you encounter projects which don't consider this a virtue, you quickly realise why it's important!).

ptersilie:
Well in that case, given all of Eco uses double quotes so far, @jaheba will have to change this.

jaheba:
I close this PR, since the commits are now merged to #88.

PR #86, opened 2016-03-11
Override selected text on paste

### LINKED ISSUE #74: override selected text
When text is selected, it should be overwritten when typing.

ptersilie:
Fixes #74


jaheba:
looks good

PR #87, opened 2016-03-14
Fix #80 No deselect on simple shift press.

### LINKED ISSUE #80: shift deletes current selection

jaheba:


jaheba:
Will be done in #81.

PR #88, opened 2016-03-14
KeyPress Refactoring

jaheba:
This PR adds to things:

A new `KeyPress` class, that wraps key presses which also supports os x.

Modification of the API for `key_cursors` and `control_cursors`. They now take a `KeyPress` object, instead of a string.

There is still some work, but I wanted to have this code as a discussion base.

### DIFF
--- a/lib/eco/treemanager.py
+++ b/lib/eco/treemanager.py
@@ -1030,7 +1030,7 @@ class TreeManager(object):
             self.unselect()
 
     def jump_cursor_within_selection(self, key):
-        '''
+        """
             Jump cursor with respect to text selection.
 
             There are 4*2 = 8 different cases, with four different
@@ -1043,7 +1043,7 @@ class TreeManager(object):
             * RIGHT: Jump to right end of selection.
             * UP:    Jump one line upwards w.r.t. left end of selection.
             * DOWN:  Jump one line downwards w.r.t. right end of selection.
-        '''
+        """
         selection_start, selection_end = sorted(
             [self.selection_start, self.selection_end])
 
@@ -1080,6 +1080,14 @@ class TreeManager(object):
         self.selection_end.pos = len(self.selection_end.node.symbol.name)
         self.cursor.pos = self.selection_end.pos
         self.cursor.node = self.selection_end.node
+
+    def select_all(self):
+        self.selection_start = Cursor(self.get_bos(), -1, 0)
+        self.cursor.node = self.get_eos()
+        self.cursor.jump_left() # for now ignore invisible nodes
+        self.cursor.pos = len(self.cursor.node.symbol.name)
+        self.cursor.line = len(self.lines) - 1
+        self.selection_end = self.cursor.copy()
 
     def unselect(self):
         self.selection_start = self.cursor.copy()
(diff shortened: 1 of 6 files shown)


ltratt:
[I'll quickly remake the `"` vs `'` comment from #81.]

jaheba:
@ptersilie I've added select all, but it behaves very strange if there are more lines in the file than currently displayed.

ltratt:
From memory, the problem with the "select all" bug is that if the selection begins/ends outside of the viewport, then it may not be displayed. In fact, I can replicate this in "plain" Eco. Simply select more than one screenful's worth of text, then scroll back up a bit: the selection disappears and reappears.

I assume that's what happening is that Eco only checks the tree in the viewport to see if there is a current selection, when instead it needs to check the whole file (in some sense).

ptersilie:
I actually fixed this problem at some point. I wonder where that fix went tho as it doesn't seem to be working anymore. I'll have a look.

ptersilie:
I pushed a few fixes that solve the drawing bug when the start and end of the selection lie outside of the screen. I also fixed your `select_all` methods which forgot to set the `line` attribute. For now I also ignored invisible nodes (e.g. `INDENT`, `DEDENT`) when using `select_all` to be consistent with using the mouse to select the whole document.

jaheba:
Should we merge soon? I think we've piled up a lot of changes in this PR.

ltratt:
Yes, this PR is getting pretty unwieldy. I'm not sure we should be trying to solve all Eco issues in one PR :)

ptersilie:
Agreed. Although in this case it was a fairly big keypress refactoring and refactorings are always ugly. The `select_all` thing and maybe the bugs is something we could've solved in another PR.

Ah well. I think it's done now. I fixed the history bug. Let's squash some commits (maybe not all, for instance the last one could be important for future reference).

ltratt:
If you wanted, we could pull out the "unrelated" commits into a separate PR. I don't feel strongly about this (though if someone wants practise with `git`, this is a fun little task).

lib/eco/treemanager.py
```diff
@@ -1048,14 +1048,22 @@ def jump_cursor_within_selection(self, key):
             [self.selection_start, self.selection_end])
 
         if key.left:
-            self.cursor = selection_start.copy()
```
jaheba: Maybe it would be good to have a method like `set_to` which does these three lines in:

``` python
self.cursor.set_to(selection)
```

jaheba:
I can start and squash the commits in meaningful chunks. We then still can put some commits into a different PR.

lib/eco/treemanager.py
ptersilie: Yes that sounds like a good idea. Do you mind doing that?

ptersilie:
Okay this is looking good. Let's try and squash this into chunks and then we merge.

jaheba:
I've split it into four commits now. We still could split one commit (`Select All`) into two though.

ptersilie:
Okay this looks good now. I think we can merge this if no one has any objections.

jaheba:
Rebase first?

ptersilie:
I am happy with the current commits.

jaheba:
So, let's hit the magic button? :trumpet: :trumpet: :trumpet: :trumpet:

PR #89, opened 2016-03-14
Tool for rebinding shortcuts in Qt's ui-files.

jaheba:
In OS X some key bindings are really different from other operating systems. Prime example is that `find next` is not `F3` but instead uses `Cmd + G`.

Since most bindings are saved in a xml file called `gui.ui` which is generated by Qt Designer this xml file has to be rewritten for Mac. Shortcuts are stored in `/action/property[@name="shortcut"]/string`.

This PR adds:
- A tool which takes a `ui` file and a ruleset (json) and replaces the specified action shortcut with the new one.
- A ruleset file.
- The tool applied to `gui.ui` using the ruleset and the result saved to `gui_osx.ui`.
- A bridge in eco which loads that file on mac systems.

### DIFF
--- a/bin/qtui-rebind
+++ b/bin/qtui-rebind
@@ -1,0 +1,34 @@
+#!/usr/bin/env python
+"""
+    Simple script to rewrite Qt ui files to change shortcut bindings.
+"""
+
+import json
+
+from lxml import etree
+import click
+
+
+def substitute(tree, name, shortcut):
+    query = ('//action[@name="{name}"]'.format(name=name)
+             + '/property[@name="shortcut"]/string')
+    result = tree.xpath(query)
+    if result:
+        result[0].text = shortcut
+
+
+@click.command()
+@click.argument('input', type=click.File('rb'))
+@click.argument('output', type=click.File('wb'))
+@click.option('--rules', '-r', type=click.File('rb'), required=True)
+def cli(input, output, rules):
+    tree = etree.parse(input)
+    ruleset = json.load(rules)
+    for name, shortcut in ruleset.items():
+        substitute(tree, name, shortcut)
+    tree.write(output, encoding='UTF-8', xml_declaration=True)
+    output.write('\n')
+
+
+if __name__ == '__main__':
+    cli()
--- a/lib/eco/eco.py
+++ b/lib/eco/eco.py
@@ -53,7 +53,7 @@ from grammars.grammars import languages, newfile_langs, submenu_langs, lang_dict
 from time import time
 import os
 import math
-
+import sys
 import syntaxhighlighter
 import editor
 
@@ -64,7 +64,11 @@ from lspace_ext import view_ecodoc_in_lspace
 
 import logging
 
-Ui_MainWindow, _     = uic.loadUiType('gui/gui.ui')
+if sys.platform == "darwin" and os.path.exists('gui/gui_osx.ui'):
+    Ui_MainWindow, _     = uic.loadUiType('gui/gui_osx.ui')
+else:
+    Ui_MainWindow, _     = uic.loadUiType('gui/gui.ui')
+
 Ui_ParseTree, _      = uic.loadUiType('gui/parsetree.ui')
 Ui_StateView, _      = uic.loadUiType('gui/stateview.ui')
 Ui_About, _          = uic.loadUiType('gui/about.ui')
--- a/lib/eco/gui/gui_osx.ui
+++ b/lib/eco/gui/gui_osx.ui
@@ -1,0 +1,637 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<ui version="4.0">
+ <class>MainWindow</class>
+ <widget class="QMainWindow" name="MainWindow">
+  <property name="geometry">
+   <rect>
+    <x>0</x>
+    <y>0</y>
+    <width>920</width>
+    <height>737</height>
+   </rect>
+  </property>
+  <property name="sizePolicy">
+   <sizepolicy hsizetype="Preferred" vsizetype="Preferred">
+    <horstretch>0</horstretch>
+    <verstretch>0</verstretch>
+   </sizepolicy>
+  </property>
+  <property name="windowTitle">
+   <string>Eco - Editor for language composition</string>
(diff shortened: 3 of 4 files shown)


ltratt:
I think you're going to need to give some hint as to what this actually does (I'm completely clueless).

jaheba:
Sorry, I added a description.

bin/qtui-rebind
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env python
+"""
+    Simple script to rewrite Qt ui files to change shortcut bindings.
```
ltratt, line 3: Random indentation?

bin/qtui-rebind
```diff
@@ -0,0 +1,34 @@
...
+"""
+
+import json
+
+from lxml import etree
+import click
```
ltratt, line 9: What is `click`?

ltratt:
If I understand correctly, the user has to manually run the rebind script? If so, why not make Eco do this when it loads?

jaheba:
Mainly because it requires xpath and thus lxml and I didn't wanted to add the dependency to eco.

bin/qtui-rebind
jaheba, line 9: See: http://click.pocoo.org/5/

It is a handy cli options utility.

bin/qtui-rebind
jaheba, line 3: I like to have docstrings being indented.

ltratt:
Personally, I think I'd be OK with the lxml dependency, if it makes Mac users' miserable lives a little brighter. [Though this is probably a decision for @ptersilie.]

bin/qtui-rebind
ltratt, line 3: I don't think that's Eco's current style (for better or worse), so I suggest we follow the project's existing precedent. [Personally, I would probably make this a comment rather than a docstring, but maybe that's just me.]

jaheba, line 3: This is all about personal preference for styling, so here is my stance on it.

I prefer

``` python
"""
    This is a longer
    docstring.
"""
```

over 

``` python
"""This is a longer
docstring.
"""
```

because it makes the block of text stand out much more and thus increase recognisability and readability, especially on module level. They are both equivalent after been trimmed.

I would disagree with the comment suggestion because it is a descriptive information about what the script does - generally speaking.

ltratt, line 3: When we're working on projects, our personal stance isn't really relevant. Except in extreme circumstances we must fit in with what's already there, even if we disagree with it.

jaheba, line 3: Agreed. But as I understand this discussion it is about exchanging ideas rather than truly convincing someone of their wrong ;)

And to be picky there aren't any other module level docstrings.

Edit: Ok, I found one short one in ast_viewer

ptersilie:
I am wondering if we can't just add a settings tab where all users can change shortcuts to whatever they like (not just OSX users). Instead of modifying the gui.ui xml files, we could store the changed shortcuts together with all other Eco settings and when Eco is loaded it updates the shortcut actions with whatever it loaded from the settings. This, of course, requires Qt to be able to update hotkeys during runtime, which I presume it can, but I haven't checked yet.

The advantage is that we can integrate those settings into Eco and we also won't need the click, lxml dependencies.

ltratt:
It would be nice not to have to maintain two XML files...

ptersilie:
Agreed. It looks like changing shortcuts is as easy as executing `self.actionOpen.setShortcut(_translate("MainWindow", "Ctrl+O", None))`. So I suggest we add a settings tab in Eco, where the user can reassign shortcuts (preferabkly by just pressing the keys and Eco reads the correct shortcut combination from that). We then store the shortcuts with all other Eco settings, and when Eco loads up we simply read them from the settings and update all the actions.

ltratt:
How about we just hardcode into the source the Mac shortcuts using the above API? I don't think this is something that's worth putting a lot of UI energy into (we have bigger fish to fry).

ptersilie:
That would be the quick and dirty solution. However I don't think creating that settings tab will take more than an hour.

jaheba:
I think the problem is not so much creating the settings tab but maintaining it.

Personally I like the way sublime text handles it. First there is a config (json) file for each platform which contains the respective keybindings. Then there is also the option to define user specific bindings which override the defaults.

Nevertheless we would have to define a set of actions which are supported in both cases first. And then there is the question how much it's actually worth implementing all that if the idea is to port everything to rust in the not so far future.

ltratt:
Agreed. I do think hard coding it via the API is the way to go (low effort now and in the future!).

ptersilie:
Okay, so we only going to hard code the mac bindings in, and not provide two separate ui-files (the other problem with the ui files and the rebinding tool would be that that marks them as changed and it's only a matter of time until someone accidentally commits their changed bindings).

So no merge for this PR then?

ltratt:
We could update this PR or open a new one; it doesn't matter too much either way.

jaheba:
Couldn't we just add a config file like `~/.eco/key-bindings` which contains all the custom keybindings instead of hard coding the bindings for mac?

ltratt:
That wouldn't be version controlled.

jaheba:
What is is that we want? Only custom bindings for mac or the general possibility for users to change the key bindings?

ltratt:
Just for Mac IMHO. [Custom key bindings are not worth bothering with IMHO.]

jaheba:
I was a bit confused since Lukas mentioned the idea of having the those settings in general.

When we only want some Mac bindings than I'm with you on that.

ltratt:
I think we're all on the same page now: something simple that works for Mac (i.e. hardcoding) is sufficient (and easy to do).

ptersilie:
Yes, I thought that if we would provide custom keybindings then it would be nice to have them configurable from within Eco. But as Laurie said it might not be worth the effort, so for now hard-coded mac bindings is enough.

So yes, I think we are all on the same page now.

jaheba:
Then let's open a ticket for that feature request, close this PR and I hard
code the OS X bindings and create a new PR for that.

PR #90, opened 2016-03-15
Fix export of Python+HTML+SQL.

snim2:
Files using a `Python + HTML + SQL` grammar have been exported as text, because the if/else block in the `treemanager` which deals with exports only recognised `HTML + Python + SQL`. 

This PR fixes that, however it still won't be possible to run a `Python + HTML + SQL` file. I'm not sure whether that is intended behaviour or not.

/cc @jaheba

### DIFF
--- a/lib/eco/eco.py
+++ b/lib/eco/eco.py
@@ -1221,7 +1221,7 @@ class Window(QtGui.QMainWindow):
         self.ui.menuChange_language_box.setEnabled(enabled)
         self.ui.actionPreview.setEnabled(enabled)
         self.ui.actionInput_log.setEnabled(enabled)
-        
+
         if enabled == False:
             self.ui.actionProfile.setEnabled(enabled)
             self.ui.actionStateGraph.setEnabled(enabled)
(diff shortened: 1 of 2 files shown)


ptersilie:
I never wrote an exporter for `Python + HTML + SQL`. So far we only used this composition within `HTML + Python + SQL` for our demos. So it initially wasn't intended to be a standalone composition. Which, however, doesn't mean that an exporter for `Python + HTML + SQL` isn't useful. So thanks for adding one. I'll have a look into this.

ptersilie:
The reason this composition doesn't run, is that the exporter assumes the outer language is `HTML`. In this composition however the outer language is `Python`, so we will have to adjust the exporter or create a new one, that can deal with the outer language being Python.

ltratt:
Yes, in this case, having Python as the outer language is a subtly different thing, and is, in essence, a new composition (needing a new exporter).

snim2:
OK, I will write a new exporter and add it to this PR as soon as I've cleared all the urgent things I need to do.

ltratt:
Yep, I don't think this is a very high priority.

snim2:
Closing this as it is low priority and need a full re-write.

PR #94, opened 2016-03-17
Hardcoded shortcuts for OS X

jaheba:


ptersilie:
Is this the only shortcut for now?

jaheba:
Yes, but there should be others. For example I think `ctrl+space` has to be
changed too.

ptersilie:
As I am not using OSX, which shortcuts we need is completely up to you. :) Let me know when I should merge something.

jaheba:
I looked through all short cuts we currently support. Those two are the only ones being different in OS X.

ptersilie:
Good. Merged.

Issue #98, opened 2016-03-18
Missing JavaScript grammar features


jaheba:
- [ ] Division: `1/2`
- [ ] "#hello"
- [ ] `1.2`;
- [ ] `var _ = 3`;
- [ ] `var $ = 4`;
- [ ] `var $_$ = 5`;
- [ ] `var x = 6 > 7`;

jaheba:
The `#` in strings seems to be fixable by adding a rule like `NONSPACE = "\S"`.

ptersilie:
Ah, yes. In the other grammars we defined a matchall token that catches everything that is not defined by some other lexing rule. This is needed to allow arbitrary text within Strings and comments.

PR #99, opened 2016-03-18
Float and hex/octal/binary numbers for Py and Js.

### LINKED ISSUE #96: hex literals in javascript and python
Both Python and Javascript allow to write numbers in binary, octal and hex format:

``` python
a = 0b01
b = 0777
c = 0xc0ffe
```

### LINKED ISSUE #97: Float Regex for Python and Javascript
``` python
r"([0-9]+\.?[0-9]*|\.[0-9]+)([eE](\+|-)?[0-9]+)?"
```

should be the regular expression which is used to detect number or float literals.

jaheba:
Fixes #96
Fixes #97


ptersilie:
Looking good. I just checked with the reference Python grammar and they only produce one token (`NUMBER`) for hex, octal, decimal and binaries. However I actually like your solution better as it's much cleaner.

ptersilie:
Do you maybe also want to update the syntaxhighlighter to highlight floats, hex and binary values?

jaheba:
Done. I've also had to rearrange the rules a bit to have highlighting for `true` and `false`.

ptersilie:
Do you mean you rearranged the grammar rules or the lexing rules? Generally it is sufficient to rearrange the lexing rules to change priorities.

jaheba:
I think I mean the lexing rules. I put `literal` before `"identifier"`.

ptersilie:
Ah, yes that sounds correct.

ptersilie:
Okay, let's squash those two commits and then I merge.

jaheba:
I've added solutions for the problems I described in #98.
- Setting `NONSPACE = "\S"` solves the string issue.
- Removing the `SLASH` rule fixed division.
- Removing `DOLLAR` and `UNDERSCORE` fixed the use of the respective single letter variables.
- Adding `$` to the `identifier` rule allows multiple `$` in the name.
- `relationalExpression` only had `in` defined, but was missing `LT`, `LTEQ`, ...

ptersilie:
The `relationExpression` missing `LT`, `LTEQ`, etc, was due to some conflicts that I guess were solved when I did some major changes to the grammar. Glad to see that putting them back in didn't trigger new conflicts. :)

All in all, this PR looks fine and I'm happy to merge it once it's squashed.

jaheba:
Commits are squashed and ready to be merged.

Issue #101, opened 2016-03-21
HDPI drawing issues


ltratt:
Eco doesn't seem to draw quite properly at HDPI resolutions (I'm running at 2560x1440):

![screen1](https://cloud.githubusercontent.com/assets/20318/13919275/c72da180-ef6c-11e5-9297-91e93022dc3b.png)

The main annoyance is that the font baseline is incorrect: note the "y" (and the brackets!) spill below the language box. It looks to me like the `y` offset of the language box is incorrect.

A minor annoyance is that the start and end brackets for the language box are indeed a pixel wide -- which at this resolution means that they're almost invisible.

jaheba:
To add to this, on (at least) Mac it seems that font sizes other than multiples of 5 yields a strange rendering for selected text:

![longselection](https://cloud.githubusercontent.com/assets/133732/13925266/03dac9f8-ef88-11e5-869f-22992c06c487.jpg)

(Please note that the whole text is selected.)

ltratt:
That problem I don't see. They may be related, of course.

jaheba:
It only is visible if you have one really long token (or word).

ltratt:
I've just tried a really long token, and I don't see that particular bug. That said, the fact that I see an incorrect y offset, and you see an incorrect width suggests either we (or Qt) have a bug with fonts at high resolutions.

jaheba:
I found a fix for my issue.
The underlying problem is that the font size is measured in points and thus is in relation to inches. So setting the font size from `15` to `16` may result in the same width for a single character, say `10`. The width of 100 characters however is quite different, in my case `963` and `1068`.

Below is the code which sets the font width. If we set it to a more precise float value, the selection displays correctly.

``` python
class GlobalFont(object):
    def __init__(self, font, size):
        font = QtGui.QFont(font, size)
        self.setfont(font)

    def setfont(self, font):
        self.font = font
        self.fontm = QtGui.QFontMetrics(self.font)
        self.fontht = self.fontm.height() + 3
        self.fontwt = self.fontm.width(" ")
        #self.fontwt = self.fontm.width(" "*100)/100.0
```

The other issue may have to do with the 3 pixels offset, but that's only a wild guess.

ltratt:
It looks like the `+3` hack only affects the line height, not the y offset within that line. It was worth trying!

I think the width thing is worth making a PR for, though.

ptersilie:
Fixed in #113

PR #108, opened 2016-03-24
Fix selection width on OS X.

jaheba:
See #101.

### DIFF
--- a/lib/eco/eco.py
+++ b/lib/eco/eco.py
@@ -86,7 +86,7 @@ class GlobalFont(object):
         self.font = font
         self.fontm = QtGui.QFontMetrics(self.font)
         self.fontht = self.fontm.height() + 3
-        self.fontwt = self.fontm.width(" ")
+        self.fontwt = self.fontm.width(" "*99)/99.0
 
 class LineNumbers(QFrame):
     def __init__(self, parent=None):


ptersilie:
Can you explain again why this works? Is it a bug in Qt (in that case we might want to submit a bug report)? Theoretically `self.fontm.width(" ")` should be the font width of every character in a fixed-width font. How come this is not true for some font sizes?

ltratt:
I _think_ the point is that the width(" ") returns a width in integer points, whereas there's no particular reason a font designer needs to make sure their characters have a width that's an integer number of points. Is that correct?

ptersilie:
Ah, that actually does make sense. Although even when we tried using the same font (Monospace) on both my and @jaheba's machine, it didn't work on his mac. Unless macos does something weird to all its fonts.

ptersilie:
Then again, I think `width()` returns the size of a string in pixels how they are rendered by Qt. I'm not sure how anything can be 1.275 pixels wide.

ltratt:
I think my comment still stands, even if the width is in pixels. There's no reason a character width needs to precisely align with pixels. [For an extreme example, imagine we rendered fonts at 2pt -- the pixels are much bigger than many of the font's strokes.]

jaheba:
I quote myself here:

> The underlying problem is that the font size is measured in points and thus is in relation to inches. So setting the font size from 15 to 16 may result in the same width for a single character, say 10. The width of 100 characters however is quite different, in my case 963 and 1068.

ptersilie:
It was that quote I had issues with. ;) Especially

> font size is measured in points and thus is in relation to inches.

jaheba:
72pts is 1inch. It refers to the actual size displayed. It has nothing to do with mac doing something weird, but simple maths. As you said correctly nothing can be `1.275px` wide. So Qt has to approximate to the nearest pixel, which would be `1px` in this case. But if you have longer text Qt can render those pixels better to fit their actual size, with the result that a 100 character text may take 128px.

ptersilie:
Ah I get it now. If my text is `abc` and the font would render that at 25px, then this would make each character 8.33 pixels wide. Conversion to int makes this 8px for each character which is why the cursor (which is drawn using font.width \* x-coordinate) is off.

I'm still wondering why this only seems to be an issue on Mac and not Linux even when using the same fonts. But I'm happy with this explanation now.

ltratt:
Yes, I _think_ that's what's going on too.

snim2:
Possibly silly question, but @jaheba which version of Python are you using to run Eco?

ltratt:
@ptersilie It's possible that width(" ") returns floats on Linux / OpenBSD? I'm not sure this is hugely likely, but it may explain why neither you or I seem able to replicate this problem. It may also be that the fonts we choose to use are designed to have "perfect" widths (if that makes sense).

jaheba:
```
$ python --version
Python 2.7.10
```

jaheba:
Maybe the aliasing algorithm behaves different?

https://en.wikipedia.org/wiki/Font_rasterization

> Mac OS X's Quartz is distinguished by the use of floating-point positioning[citation needed]; it does not force glyphs into exact pixel locations, instead using various antialiasing techniques, including subpixel rendering, to position characters and lines to appear closer to the type designer's intent.

ptersilie:
@ltratt We tried using the same fonts tho (Monospace). So unless they are different on Mac, it doesn't seem the fonts being the problem.

ltratt:
It's possible.

Anyway, assuming this PR doesn't break things on Linux, I think it's worth merging.

ptersilie:
Of course. I'll run a few more checks and then merge it.

PR #109, opened 2016-03-24
Fix no reevaluation on cut or paste.

### LINKED ISSUE #95: No evaluation of tree after cutting or pasting code
When pasting code the evaluation of the tree is not updated. The same happens when new code is pasted into eco.

jaheba:
Fixes #95.


jaheba:
Fixed in #110.

PR #110, opened 2016-03-24
Trigger reparsing on pasting and cutting

### LINKED ISSUE #95: No evaluation of tree after cutting or pasting code

ptersilie:
Fixes #95.

Same as PR #109.


jaheba:
Looks good.

PR #117, opened 2016-04-13
Improve the zoom quality of the parse tree

ptersilie:


ltratt:
It took me a long time to review this because of the size of the diff.

I also tested it, and it is a definite improvement. Thanks!

PR #119, opened 2016-04-21
Fix first undo

ptersilie:
Previously when opening an eco file, then type something and immediately press undo, the undo didn't work. This fixes this.

### DIFF
--- a/lib/eco/treemanager.py
+++ b/lib/eco/treemanager.py
@@ -683,6 +683,8 @@ class TreeManager(object):
 
     def key_ctrl_z(self):
         self.log_input("key_ctrl_z")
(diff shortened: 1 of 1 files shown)


lib/eco/treemanager.py
```diff
@@ -683,6 +683,8 @@ def get_max_version(self):
 
     def key_ctrl_z(self):
         self.log_input("key_ctrl_z")
+        if not self.undo_snapshots and self.get_max_version() > 1:
```
ltratt: If `undo_snapshots` is a list, it's probably clearer to say `if len(undo_snapshots) == 0`?

ltratt:
Looks good to me. Squash and I'll commit?

ptersilie:
Squashed & ready.

PR #122, opened 2016-04-28
Jruby fix viewport error

snim2:
This also fixes a typo in Settings.

The main fix is this: previously when a railroad line (from the JRuby morphism demo) fell outside of the viewport, the whole set of lines for that method would disappear. This is how the demo looked before the bug fix (on the current master branch):

[Before bug fix video](https://youtu.be/nU5pzSeyFZw)

This is how the morphism demo looks with this bug fix: 

[After bug fix video](https://youtu.be/83Wn_0YMQJg)

### DIFF
--- a/lib/eco/overlay.py
+++ b/lib/eco/overlay.py
@@ -76,6 +76,11 @@ class Overlay(QWidget):
         for key in datum:
             self._railroad_data[key] = datum[key]
 
+    def add_railroad_data(self, data):
+        for datum in data:
+            for key in datum:
+                self._railroad_data[key] = datum[key]
+
     def clear_data(self):
         self._heatmap_data = dict()
         self._railroad_data = dict()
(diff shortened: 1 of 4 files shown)


lib/eco/overlay.py
```diff
@@ -76,10 +76,30 @@ def add_railroad_datum(self, datum):
...
+        total_height = 0.0
+        gfont = QApplication.instance().gfont  # Current system font.
+        for line_no, line in enumerate(self.tm.lines):
+            if (line_no + 1) > start_line:
+                end_line += 1
+                total_height += line.height * (gfont.fontht)
+            if total_height > viewport_height:
+                break
+        return (start_line, end_line)
+
```
ptersilie: This looks correct. I'm not 100% sure but you might be able to get `end_line` from `self.paint_nodes` which returns the last line drawn. Then you won't have to compute this number again. This of course requires that the overlay is drawn _after_ the source code has been painted.

snim2:
OK, I think this is the change you are after. Let me know if you are happy with that and I will squash.

ptersilie:
Oh great. It worked. Good, if you squash I will merge.

snim2:
Done!

PR #125, opened 2016-05-05
Another attempt at fixing the default icon theme

ptersilie:


vext01:
Can you explain the logic behind the change? Is it that not all icon sets have "document-new", but gnome does, and it's a commonly installed icon set?

ptersilie:
Our first attempt didn't work because `themeName` doesn't necessarily have to be set for icons to work (icon themes can be defined by the OS as well, in which case `themeName` returns None). So I changed the check to see that when we request an icon, something is actually returned. I chose `document-new` because it's an icon that I assume most icon sets will have. If no icon was loaded that means that no icon theme is defined in which case we default to gnome, which is a commonly installed icon set.

This probably doesn't catch all cases, but it's better than having no fallback at all.

vext01:
@snim2 does this work for you?

snim2:
It seems fine with some themes (e.g. Faenza) but with others (e.g. Gnome) I don't see any icons at all.

ptersilie:
Interesting. I don't have any problems with it, no matter which icon theme I choose in my window manager. But I probably never get the fallback because I have always something selected.

vext01:
So I guess we need to find a better fix so it works under the gnome theme on sarah's machine. One part of me, however, thinks we shouldn't spend too much time on this...

ltratt:
I tend to agree with Edd. I think we're getting close to "doctor it hurts everytime I do X" "well don't do X" territory.

One possibility is for Sarah to work out what the magic sequence of:

``` python
if <desktop> == "gnome":
   icon_theme = "whatever"
else:
  icon_theme = "document-new"
```

is. It's plausible that there simply isn't a one-size-fits-all option available to us.

ptersilie:
Agreed. I just thought it was better to have this fix rather than none, but I don't feel I want to spent much more time on this, given that in the end this should be configured properly in the window manager, rather than Eco itself.

vext01:
Shall we merge?

snim2:
+1

ptersilie:
Yes please. Because the current fix always falls back to gnome for me (and probably everone else who has their own icon theme set).

PR #129, opened 2016-05-25
Draw callgraph visualisation over polyglot code

snim2:
This PR has been tested with JRuby+JavaScript and JRuby+SimpleLanguage code. Example screenshots are shown below. You may notice that code within the language boxes (i.e. code not written in Ruby) is not visualised. This is because the JS and SL Truffle interpreters do not add the necessary information to the callgraph log.

polyglot_callgraph_js
polyglot_callgraph_sl

### DIFF
--- a/lib/eco/export/jruby.py
+++ b/lib/eco/export/jruby.py
@@ -219,39 +219,16 @@ class JRubyExporter(object):
                                 stderr=subprocess.STDOUT,
                                 bufsize=0)
 
-    def _annotate_text(self, lineno, text, annotation, klass):
-        """Annotate a node on a given line with a given symbol name.
-        FIXME: Should handle case where line (and all subsequent lines) have
-        already been deleted by the user.
-        FIXME: Needs to handle the case where same text appears in more than
-        one node on the same line.
-        """
-        try:
-            temp_cursor = self.tm.cursor.copy()
-            temp_cursor.line = lineno - 1
-            temp_cursor.move_to_x(0)
-            node = temp_cursor.find_next_visible(temp_cursor.node)
-            while node.lookup == '<ws>' or node.symbol.name != text:
-                node = node.next_term
-                if isinstance(node, EOS):
-                    raise ValueError('EOS')
-            node.remove_annotations_by_class(klass)
-            node.add_annotation(klass(annotation))
-        except ValueError:
-            pass
-
     def _profile(self):
-        """Run JRuby and dump a callgraph to disk.
-        Parse the callgraph and annotate the syntax tree.
-        """
+        callgraph_processor = JRubyCallgraphProcessor(self.tm)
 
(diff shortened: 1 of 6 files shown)


ltratt:
Why are line numbers possibly (always?) inaccurate in polyglot programs? [I ask this naively -- I have no idea of the particular technical limitations in this case.]

snim2:
When polyglot code is exported, Eco has to add some boilerplate to the
original program. This exposes the JS or SL code to JRuby.

Also, between the user pressing 'Profile' and the visualisation appearing
on screen, there is time for the user to edit their code, which may make
the line numbers in the JRuby callgraph invalid.

lib/eco/export/jruby.py
```diff
@@ -381,3 +389,63 @@ def _profile(self):
...
+                            self._annotate_text(callsite_version.callsite.line,
+                                                method.name,
+                                                { method.name : method.is_mega },
+                                                JRubyMorphismLine)
+                        num_calls += 1
+
+    def _annotate_text(self, lineno, text, annotation, klass):
+        """Annotate a node on a given line with a given symbol name.
+        """
+        if lineno < 0:
```
ltratt: How can `lineno < 0` be true?

lib/eco/export/jruby.py
```diff
@@ -381,3 +389,63 @@ def _profile(self):
...
+            return
+        # Attempt to find 'text' on 'lineno'.
```
ltratt, line 416: I'm not saying that there's a nicer approach than this, but I wonder if having so much code in the `try` block is a good plan?

ltratt:
I've made a couple of minor comments and have one over-arching thing. Can we be consistent about using `"` or `'` for strings? I think the Eco codebase uses the former, but I don't care, as long as we're consistent.

ltratt:
[I'm also not quite sure about formatting function arguments with each on its own line, but that's a slightly harder thing to make consistent easily.]

lib/eco/export/jruby.py
snim2: This happens when method is called which has not been defined by the programmer. For example,  in the snippet:

``` ruby
10.times do
    puts "hello"
end
```

`times` is a method on the object `10`, but the user has not defined `times`, so according to the callgraph log, the source section from which `times` is defined starts on line -1. Below is an example of (part of) a callgraph from one of my examples, you can see there are two examples in that digraph of methods defined on line -1:

test_times_callgraph

ltratt: Aha, OK.

lib/eco/export/jruby.py
snim2, line 416: It isn't ideal, but I'm trying to avoid searching the whole tree, where possible.

snim2:
OK, quotes are now consistent. Ready to squash?

PR #134, opened 2016-06-15
New Ruby parser

snim2:
Use an external program to parse Ruby programs.

Fixes #120

PR #135, opened 2016-06-16
Improve settings

### LINKED ISSUE #133: Improve open file / directory dialogs in Settings
When a user changes their settings, the current behaviour of all open file / directory dialogs is to open in the CWD. If the user has already bound a setting to `some/very/very/very/very/long/path/executable` it would be nicer to open the file / directory dialog directory `some/very/very/very/very/long/path/`

snim2:
This PR improves the settings dialog. Previously, the file / directory open dialog always opened in the CWD. Following this PR, the dialog will check whether the relevant setting already has a value set, and if it does, will open in the same directory as the current setting.

If the relevant setting has was not set, the dialog opens in the `$HOME` directory.

This PR also adds the ruby-parser to the Settings dialog.

Fixes #133 
Fixes an issue discussed in #134

### DIFF
--- a/lib/eco/rubyparser/rubyparser.py
+++ b/lib/eco/rubyparser/rubyparser.py
@@ -1,8 +1,9 @@
-from grammars.grammars import EcoFile
 from export.jruby import JRubyExporter
 from incparser.astree import AST, BOS, EOS, TextNode
-from grammar_parser.gparser import Terminal, Nonterminal
 from incparser.syntaxtable import FinishSymbol
+from grammars.grammars import EcoFile
+from grammar_parser.gparser import Terminal, Nonterminal
+from PyQt4.QtCore import QSettings
 
 import subprocess
 import tempfile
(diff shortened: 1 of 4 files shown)


snim2:
I think that last build failed due to the "randomised" tests, so I will restart it and see what happens.

lib/eco/rubyparser/rubyparser.py
```diff
@@ -115,18 +116,22 @@ def _find_node(self, lineno, charno):
                 return node
 
     def inc_parse(self, line_indents = [], reparse=False):
+        settings = QSettings("softdev", "Eco")
+        ruby_parser =str(settings.value("env_ruby_parser", "").toString())
```
ltratt: Space missing after `=`.

ltratt:
I wonder if we want to say "Ruby parser" rather than "JRuby parser"? I don't think there's anything JRuby specific here is there?

snim2:
That means creating a new pane on the Settings dialog. Can do.

ltratt:
Maybe we could have a Ruby pane with a JRuby sub-part of it?

snim2:
Ready to squash?

New settings pane:

new_ruby_settings

ltratt:
I don't get all of the icons on my machine (which isn't worth worrying about), but otherwise this looks good. Squash away!

snim2:
All squashed!

snim2:
Thanks

PR #137, opened 2016-06-20
Removed unused class LineNumbers from eco.py

djmcalleja:
The LineNumbers class that is actually used is in editortab.py, so I deleted the unused one.


ptersilie:
Ah, yes, this is a left-over from when I moved this class to `editortab.py`. While we are at it we could also remove `ScopeScrollArea` from `eco.py` as this has also been moved to `editortab.py`.

PR #140, opened 2016-06-22
Remove dead code

snim2:
This PR:
- Removes some dead code.
- Removes code which has been commented. 
- Fixes some indentation. 
- Improves some spelling.

### DIFF
--- a/lib/eco/eco.py
+++ b/lib/eco/eco.py
@@ -20,7 +20,7 @@
 # IN THE SOFTWARE.
 
 from __future__ import print_function
-import subprocess, sys
+import sys
 
 try:
     import py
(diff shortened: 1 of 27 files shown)


lib/eco/eco.py
```diff
@@ -466,6 +450,8 @@ def change(self, index):
             self.ui.textEdit.setText(text)
         elif index == "Default":
             text = self.tm.export("/dev/null")
+            if text is None:
+                text = ""
```
ptersilie, line 454: Are you sure this can be deleted? I think if export fails it returns None and setText might not work with None as the argument.

snim2, line 454: These were two lines that I added, because I saw a traceback when I was testing these changes. Perhaps this should have been a separate PR?

ptersilie, line 454: Ah sorry, this is what you added. That's correct then. :) My bad.

ptersilie:
What a mess. Thanks for cleaning up a bit. Looks good.

PR #141, opened 2016-06-27
Python debugger

djmcalleja:
Added debugger for Python. Need remote pdb installed to work. 
With a Python Eco file open, start the debugger by clicking "Project -> Run with debugger" from the main menu or click the button with the trash icon next to the run button.
- Can step through code with the two buttons at the top right of the window.
- Set normal breakpoints by double clicking a line number. Set a temporary breakpoint by right clicking the line number and selecting the second option. If you right click a line number and select the third option, you can set conditional breakpoints, and set how many times to ignore the breakpoint. Delete a breakpoint by double clicking the line number.
- Check the value of a Python expression in the context of wherever the debugger is by entering it in the box under the console.
- To continue execution until next breakpoint or the program ends, click the play button near the top right. Click the stop button to quit the debugger.


snim2:
This is nice! I haven't looked through the code yet, but trying this out I notice that if I set a breakpoint and then click "run" the debugger, the debugger does not stop at the breakpoint.

Also, if I close Eco whilst the debugger is running, rpdb doesn't get killed off. So, next time I start Eco I can't debug anything. So, the GUI thread needs to communicate that it is shutting down to the debugger thread. I think the profilers might do this, so it might be worth looking at that code.

lib/eco/eco.py
```diff
@@ -1338,6 +1529,119 @@ def run(self):
                         line = "  " + line[9:]
                 self.emit(self.signal, line.rstrip())
 
+class Debugger(QObject):
+    def __init__(self, window):
```
snim2: Because the eco.py file is already very long, it would be nice to factor this class out into its own file.

lib/eco/editortab.py
```diff
@@ -204,35 +224,108 @@ def decVSlider(self, page=False):
...
+            line += 1
+        start = editor.paint_start[0]
+        return line+start+1
 
     def paintEvent(self, event):
         gfont = QApplication.instance().gfont
-        paint = QtGui.QPainter()
-        paint.begin(self)
+        paint = QtGui.QPainter()            
+        paint.begin(self)
```
snim2: Extra space at the end of some lines here. You can probably get your editor to delete this stuff automatically on save.

lib/eco/editortab.py
```diff
@@ -204,35 +224,108 @@ def decVSlider(self, page=False):
...
+            paint.setBrush(QColor("yellow")) 
+            if move:                
+                paint.drawEllipse(small_rect)  
+            else:              
+                paint.drawEllipse(big_rect)      
+
     def update(self):
-        gfont = QApplication.instance().gfont
+        gfont = QApplication.instance().gfont        
         editor = self.parent().editor
```
snim2, line 313: Space at end of line here too.

snim2:
Generally this is looking really nice! When you and @ptersilie have finished with the new changes I'll go through a more detailed code review.

lib/eco/export/cpython.py
```diff
@@ -59,6 +61,30 @@ def _run(self):
         self.tm.export_as_text(f[1])
         return subprocess.Popen(["python2", f[1]], stdout=subprocess.PIPE, stderr=subprocess.STDOUT, bufsize=0)
 
+    def _debug(self):
+        f = tempfile.mkstemp(suffix='.py')        
+        code = self.tm.export_as_text(f[1])
+
```
ptersilie, line 68: I think it would be useful here (or wherever it makes more sense) to check if the user has `remote_pdb` installed, and if not throw a warning. Otherwise the user gets a traceback.

lib/eco/debug.py
```diff
@@ -0,0 +1,148 @@
+from time import sleep
+import telnetlib
+
+from PyQt4 import QtCore
+from PyQt4.QtCore import QObject
```
snim2: You can just say `from PyQt4.QtCore import QObject` without the line above.

lib/eco/debug.py
```diff
@@ -0,0 +1,148 @@
...
+from fcntl import fcntl, F_GETFL, F_SETFL
+from os import O_NONBLOCK, read
+from socket import error as socket_error
+
+class Debugger(QObject):
+    def __init__(self, window):
+        QObject.__init__(self)
+        self.window = window
+        # used from outside debugger
```
snim2: Maybe change these comments to something like "incoming signals" / "outgoing signals".

lib/eco/debug.py
```diff
@@ -0,0 +1,148 @@
...
+        self.signal_command = QtCore.SIGNAL("command")
+        self.signal_start = QtCore.SIGNAL("start")
+        # signals that will be emitted from debugger
+        self.signal_done = QtCore.SIGNAL("finished")
+        self.signal_output = QtCore.SIGNAL("output")
+        self.signal_toggle_buttons = QtCore.SIGNAL("disablebuttons")
+        self.tn = None
```
snim2: `tn` is a bit cryptic, is this a telnet session?

djmcalleja: Yes

lib/eco/eco.py
```diff
@@ -1265,8 +1336,121 @@ def toggle_menu(self, enabled=True):
 
         if enabled == False:
             self.ui.actionProfile.setEnabled(enabled)
+            self.ui.actionDebug.setEnabled(enabled)
             self.ui.actionStateGraph.setEnabled(enabled)
 
+    def debug_continue(self):
+        # pdb Command
```
snim2: Since all these methods are called `debug_.*`  I think you can delete the comment `# pdb command`

lib/eco/eco.py
```diff
@@ -1265,8 +1336,121 @@ def toggle_menu(self, enabled=True):
...
+            removed_same_type = (isTemp == False)
+            if not bp_number:
+                bp_number = self.debugger.get_breakpoints(True, number, False)  
+                removed_same_type = (isTemp == True)      
+            if bp_number:
+                # breakpoint exists, delete it
+                self.debugger.run_command("cl " + str(bp_number))
+                removed_bp = True
+                output = ""
+            if not ((removed_bp and removed_same_type) or (from_click and removed_bp)) or from_dialog:
```
snim2: Is this line over 80 characters?

lib/eco/eco.py
```diff
@@ -1265,8 +1336,121 @@ def toggle_menu(self, enabled=True):
...
+                self.debugger.run_command("ignore " + bp_index + " " + str(ignore))
+
+    def debug_breakpoint_condition(self, number):
+        # Open window to create breakpoint
+        self.breakpointdialog.show()
+        self.breakpointdialog.setWindowTitle("Set Breakpoint on line "+str(number))
+        self.breakpointdialog.line_number = number
+
+    def debug_expression(self):
+        #pdb Command
```
snim2: likewise...

lib/eco/eco.py
```diff
@@ -1265,8 +1336,121 @@ def toggle_menu(self, enabled=True):
...
+        if previous:
+           if self.expression_num+1 < len(self.expression_list):
+               self.expression_num += 1          
+        else:
+           if self.expression_num-1 >= 0:
+                self.expression_num -= 1
+        
+        self.ui.expressionBox.setText(self.expression_list[self.expression_num])
+            
+
```
snim2: One blank line too many here

lib/eco/editortab.py
```diff
@@ -204,6 +224,52 @@ def decVSlider(self, page=False):
...
+        if line_clicked > len(editor.lines):            
+           return None;
+
+        event.accept()
+        menu = QMenu(self)
+        bAction = menu.addAction("Toggle breakpoint at "+str(line_clicked))  
+        tbAction = menu.addAction("Toggle temp breakpoint at "+str(line_clicked))
+
+        # Only offer the condition action if breakpoint exists?
+        #bcAction = False
```
snim2: Let's remove these commented lines of code, and the ones below

snim2:
This is really good!

Just a couple of general comments. It is preferable to avoid variable names with one or two letters, better to have `breakpoint` than `bp`. I realise that Eco is not consistent about this, but it does make the code more readable.

Secondly you have a few methods and functions that look like this:

```
def function(args):
    # comment here
```

In Python we usually write function or method-level comments like this:

```
def function(args):
    """comment here
    """
```

So it would be good to make that consistent throughout.

Other than those very minor points, this all looks fantastic. When you have dealt with the comments above, you should squash some commits here (ping me on MM if you haven't done this before), and then we will need to bring this up to date with master. After that, I'll merge.

lib/eco/eco.py
```diff
@@ -484,6 +489,27 @@ def focus(self):
...
 
+class BreakpointDialog(QtGui.QDialog):
+    def __init__(self, parent):
+        QtGui.QMainWindow.__init__(self)
+        self.parent = parent
+        QtGui.QDialog.__init__(self)
+        self.ui = Ui_Breakpoint()
+        self.ui.setupUi(self)
+        self.line_number = 0
+
```
snim2: We can lose this blank line

snim2:
Great! This looks ready to squash!

PR #146, opened 2016-06-30
FIX code to walk JRuby callgraph.

snim2:
This PR is the second in a series of changes working towards extending the JRuby callgraph demo to deal with programs in multiple files (the first PR was #135)

This PR fixes a bug in the JRuby callgraph demo. The bug prevented the callgraph walker from knowing which files each method was defined and called in, however, was not visible in the old videos and screenshots because we worked around it.

As an example, you can see from this callgraph (derived from the same code) that callsites and definitions can now be shown to have different source files:

callgraph_multifile

### DIFF
--- a/lib/eco/export/jruby.py
+++ b/lib/eco/export/jruby.py
@@ -187,8 +187,8 @@ class JRubyExporter(object):
         if run:
             return self._run()
         elif profile:
-            return self._profile()
-        if path is not None:
+            return self._profile(path=path)
+        elif path is not None:
             self._export_as_text(path)
             return
 
@@ -230,7 +230,7 @@ class JRubyExporter(object):
                                 stderr=subprocess.STDOUT,
                                 bufsize=0)
 
-    def _profile(self):
+    def _profile(self, path):
         callgraph_processor = JRubyCallgraphProcessor(self.tm)
 
         _, src_file_name = tempfile.mkstemp(suffix=".rb")
@@ -241,12 +241,16 @@ class JRubyExporter(object):
                                      next(tempfile._get_candidate_names()) + ".txt")
         logging.debug("Placing callgraph trace in", log_file_name)
 
-        # Run this command:
-        #  $ jruby -X+T -Xtruffle.callgraph=true -Xtruffle.callgraph.write=test.txt -Xtruffle.dispatch.cache=2 FILE
         settings = QSettings("softdev", "Eco")
         jruby_bin = str(settings.value("env_jruby", "").toString())
+        directory = str(settings.value("env_jruby_load", "").toString())
+        if directory:
+            load_path = "-I" + directory
+        else:
+            load_path = ""
         pic_size = str(settings.value("graalvm_pic_size", "").toString())
         cmd = [jruby_bin, "-X+T", "-Xtruffle.callgraph=true",
+               load_path,
                "-Xtruffle.callgraph.write=" + log_file_name,
                "-Xtruffle.dispatch.cache=" + pic_size,
                src_file_name]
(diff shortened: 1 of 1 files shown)


lib/eco/export/jruby.py
```diff
@@ -241,19 +241,21 @@ def _profile(self):
                                      next(tempfile._get_candidate_names()) + ".txt")
         logging.debug("Placing callgraph trace in", log_file_name)
 
-        # Run this command:
-        #  $ jruby -X+T -Xtruffle.callgraph=true -Xtruffle.callgraph.write=test.txt -Xtruffle.dispatch.cache=2 FILE
         settings = QSettings("softdev", "Eco")
         jruby_bin = str(settings.value("env_jruby", "").toString())
+        directory = str(settings.value("env_jruby_load", "").toString())
+        load_path = "-I" + directory if directory else ""
```
ltratt: I find Python's ternary operator a bug-inducing disaster. Can we make it a standard if/then/else? Thanks!

lib/eco/export/jruby.py
```diff
@@ -365,49 +367,74 @@ def annotate_tree(self, src_file_name, log_file_name):
 
         # Process graph of reachable objects and annotate parse tree.
         self.remove_all_annotations()
+
         for obj in reachable_objects:
-            if (isinstance(obj, Method) and obj.source.file != "(unknown)"
-                and not obj.is_hidden() and not obj.is_core()
-                and not obj.is_library() and obj.source.file != "(eval)"
-                and obj.name != "<main>"):
+            if (isinstance(obj, Method) and obj.source.file != "(unknown)" and
```
ltratt, line 375: Not a fixable comment in this PR, but I can't help feel that `obj.source.file` being "(unknown)" isn't a great way of signifying a special status!

lib/eco/export/jruby.py
```diff
@@ -365,49 +367,74 @@ def annotate_tree(self, src_file_name, log_file_name):
...
                 # Find and annotate method calls.
                 for version in method.versions:
-                    for callsite_version in version.called_from:
+                    if len(version.called_from) == 0:
+                        continue
+                    for caller in version.called_from:
+                        call_filename = caller.callsite.method.source.file
+                        # For now, if a method was called in the currently
+                        # open file, but defined elsewhere, we ignore it.
+                        if ((call_filename != src_file_name) or
```
ltratt: One level of brackets too many?

ltratt:
A few minor comments, but otherwise looks good!

lib/eco/export/jruby.py
snim2, line 375: I'm not sure what the alternatives would be...

snim2:
Needs squashing...

lib/eco/export/jruby.py
ltratt, line 375: Nor am I, but it might be worth thinking about at some point. It just seems a bit prone to subtle breakage in the future.

snim2, line 375: AIUI the "(unknown)" is inserted by the interpreter when a method has been automatically generated, and not written into the code by the programmer. So, we are safe until the callgraph log changes its syntax, in which case we will no doubt have to make further changes anyway.

ltratt, line 375: Ah, I see!

snim2:
Ugh, GitHub isn't showing me which lines of code you are referring to :(

snim2:
I might be losing my marbles but I can't see any! Which line are you looking at?

ltratt:
Click on "f5e3135" in the "ltratt commented on" bit and you should see. If not, let me know.

snim2:
OK, do you feel this is ready for squashing?

ltratt:
Yep.

snim2:
Splat.

snim2:
Thanks.

PR #147, opened 2016-07-01
Pygameviewer

ptersilie:
This PR adds a new treeviewer to Eco, based on rpythons dotviewer.

Things that need to be checked when reviewing this PR are:
- [x] Toggling whitespace on/off (Key: W)
- [x] Moving through the tree history (Key: Meta left/right)
- [x] Closing the treeviewer and reopening it without crashing it

### DIFF
--- a/lib/eco/incparser/astree.py
+++ b/lib/eco/incparser/astree.py
@@ -262,6 +262,17 @@ class Node(object):
                 return
             version -= 1
 
+    def get_attr(self, attr, version):
+        if version is None:
+            return self.__getattribute__(attr)
+        version = int(version)
+        while version >= 0:
+            try:
+                return self.log[(attr, version)]
+            except KeyError:
+                version -= 1
+        raise AttributeError("Attribute %s for version %s not found." % (attr, version))
+
     def remove_child(self, child):
         for i in xrange(len(self.children)):
             if self.children[i] is child:
--- a/lib/eco/pgviewer.py
+++ b/lib/eco/pgviewer.py
@@ -1,0 +1,181 @@
+import re
+import pydot
+from dotviewer import drawgraph
+from grammar_parser.gparser import MagicTerminal
+from threading import Thread
+from grammar_parser.bootstrap import AstNode, ListNode
+
+class PGThread(Thread):
+
+    def __init__(self, gl):
+        Thread.__init__(self)
+        self.gl = gl
+
+    def run(self):
+        self.gl.display()
+
+    def quit(self):
+        drawgraph.display_async_quit()
+
+class PGViewer(object):
+
+    def __init__(self, root, version):
+        self.root = root
+
+        self.whitespaces = True
+        self.showast = False
+        self.version = version
+        self.restrict_nodes = []
+        self.ast = None
+        self.links = {}
+        self.gl = drawgraph.GraphLayout(0,0,0)
+        self.gl.pgviewer = self
+        self.max_version = 0
+        self.t = None
+
+    def quit(self):
+        self.t.quit()
+
+    def run(self):
+        self.create_graph()
+        self.t = PGThread(self.gl)
+        self.t.start()
+
+    def is_alive(self):
+        if not self.t:
+            return False
+        return self.t.is_alive()
+
+    def send(self, command, *args):
+        if command == "refresh":
+            self.refresh(self.version + args[0])
+        elif command == "toggle_ws":
+            self.whitespaces = self.whitespaces ^ True
+            self.refresh(self.version)
+        elif command == "toggle_ast":
+            self.showast = self.showast ^ True
+            self.refresh(self.version)
+
+    def refresh(self, version):
+        if self.is_alive():
+            if version > 0 and version <= self.max_version:
+                self.version = version
+                self.create_graph()
+                self.gl.gdisplay.zoom(self.gl.scale)
+                self.gl.gdisplay.updated_viewer()
+                self.gl.gdisplay.redraw_now()
+
+    def create_graph(self):
+        self.links = {}
+        self.gl.reset()
+        graph = pydot.Dot(graph_type='digraph')
+        if self.showast:
+            self.add_node_ast(self.root, graph)
+        else:
+            self.add_node(self.root, graph)
+
+        lines = graph.create_dot("plain").splitlines(True)
+
+        _, scale, width, height = lines[0].strip("\n").split(" ")
+        self.gl.scale = float(scale)
+        self.gl.boundingbox = float(width), float(height)
+        self.gl.links = self.links
+
+        for line in lines[1:]:
+            args = splitline(line)
+            if args[0] == "node":
+                self.gl.add_node(*args[1:])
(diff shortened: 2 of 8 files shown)


snim2:
BTW #141 is right in the middle of a really tricky set of merge conflicts, to bring that branch up to date with the current master. If you could wait until PR #141 has been merged before merging this PR, that would be really helpful.

ptersilie:
@snim2 sure can do. How long until you are merging #141? As I need the treeviewer PR before I can continue on my other branches.

snim2:
@ptersilie much appreciated. Work on the branch is finished and squashed, it just needs bringing up to date before I can merge it. I'm expecting to be able to do that today (but the merge conflicts are tricky).

lib/eco/incparser/astree.py
```diff
@@ -262,6 +262,18 @@ def load(self, version):
...
+        if version is None:
+            return self.__getattribute__(attr)
+        version = int(version)
+        while version >= 0:
+            try:
+                return self.log[(attr, version)]
+            except KeyError:
+                version -= 1
+        raise Exception("Warning: reached zero", self, self.log)
+        return None
```
cfbolz: the `return None` is dead code. Shouldn't this be a somewhat less general exception type?

ptersilie: Ah, yes you are right. I'm actually not even sure what I want to do here, as in if I even want an exception here. I guess I go with a exception for now. Do you have a suggestion for a better exception?

lib/eco/pgviewer.py
```diff
@@ -0,0 +1,135 @@
...
+        if lookup == "<ws>":
+            label = "<ws> x%s" % len(label)
+        label = label.replace("\"", "\\\"")
+        label = label.replace("\\\\\"", "\\\\\\\"")
+        label = label.replace("\r", "\\n")
+        dotnode = pydot.Node("\"%s\"" % id(node), label='"%s"' % label, shape="box")
+
+        try:
+            changed = node.get_attr("changed", self.version)
+        except Exception:
```
cfbolz: catching an ignoring every kind of exception here hides errors, just catch a precise type.

lib/eco/pgviewer.py
```diff
@@ -0,0 +1,135 @@
...
+            c_node = self.add_node(node.symbol.parser, graph)
+            graph.add_edge(pydot.Edge(dotnode, c_node))
+
+        return dotnode
+
+def splitline(line, re_word = re.compile(r'[^\s"]\S*|["]["]|["].*?[^\\]["]')):
+    result = []
+    for word in re_word.findall(line):
+        if word.startswith('"'):
+            word = eval(word)
```
cfbolz: please don't use eval on user input ;-)

cfbolz:
All the three things you wanted me to test work.

cfbolz:
Maybe you can add the eco-specific key bindings to the in-viewer help (press h or ? to get it).

lib/eco/pgviewer.py
ptersilie: This is not my code. ;)

lib/eco/pgviewer.py
```diff
@@ -0,0 +1,135 @@
...
+        except Exception:
+            label = node.symbol.name
+            lookup = node.lookup
+        if label == "\r":
+            label = "<return>"
+        if lookup == "<ws>":
+            label = "<ws> x%s" % len(label)
+        label = label.replace("\"", "\\\"")
+        label = label.replace("\\\\\"", "\\\\\\\"")
+        label = label.replace("\r", "\\n")
```
cfbolz: this looks strange, do you still need it?

snim2:
PR #141 has been merged now, thanks for waiting.

ptersilie:
I already added the keybindings to the help.

ptersilie:
Merged master into branch

ltratt:
I think we can (and should) fix the `eval` issue by doing something like:

``` python
if word.startswith("\""):
  assert word[-1] == "\""
  word = word[1:-1]
```

ptersilie:
If that is all this `eval` does, then yes.

ltratt:
Please check, but I _think_ that's purely what it's aiming to do. [Of course, it can also eval any code, so you could pass `"blah" + os.system('rm -r /')` and it would happily wipe your hard disk!]

ptersilie:
Seems to be working and I agree with your fix.

ltratt:
OK, excellent!

One thing I'm unconvinced by is including a font: DroidSans is widely available (even I have it installed!) and we should just tell people to install it. If you feel really strongly about using that particular font, you can even print a warning saying "DroidSans not found: install it from X", but I don't think we want to start shipping ttf files.

ptersilie:
I didn't write this code. I don't feel strongly about anything. ;) That being said I agree that not shipping a font is probably a good idea.

ltratt:
Hehe :) OK, let's just delete it, and not worry about a warning then.

ptersilie:
Removed the fonts and added checks to see if droid is installed and if not fall back to the default system font

ltratt:
If I press "r" for "reload" I get:

```
  File "/usr/local/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/home/ltratt/src/eco/lib/eco/pgviewer.py", line 15, in run
    self.gl.display()
  File "/home/ltratt/src/eco/lib/eco/dotviewer/drawgraph.py", line 81, in display
    self.get_display().run()
  File "/home/ltratt/src/eco/lib/eco/dotviewer/graphdisplay.py", line 749, in run
    self.run1()
  File "/home/ltratt/src/eco/lib/eco/dotviewer/graphdisplay.py", line 743, in run1
    self.process_event(EventQueue.pop(0))
  File "/home/ltratt/src/eco/lib/eco/dotviewer/graphdisplay.py", line 631, in process_event
    method(event)
  File "/home/ltratt/src/eco/lib/eco/dotviewer/graphdisplay.py", line 690, in process_KeyDown
    method(*args)
  File "/home/ltratt/src/eco/lib/eco/dotviewer/graphdisplay.py", line 431, in reload
    self.layout.request_reload()
AttributeError: GraphLayout instance has no attribute 'request_reload'
```

ltratt:
If I load in `rng.eco`, I get one language box's root node in green, but no other node is coloured the same way. Is this significant?

ptersilie:
The green thing means the node has changes. I never clear up the changed flag of the root (which I probably should out of consistency) but it never matter as the root is always changed anyways.

ltratt:
In this case, the root root (if that makes sense) isn't green, but some random language box's root is green.

ptersilie:
Yes, it doesn't matter at the moment that some root nodes are changed. For consistency reasons they shouldn't be, but since we always go down that path it doesn't change anything either. This is also something in the parser that has nothing to do with this PR.

ptersilie:
I fixed the traceback on reload issue tho

ltratt:
OK, fair enough. So once the tests have passed, let's squash, then I think we can merge.

PR #149, opened 2016-07-05
copy over revision 7f5d1ceedd1dffa5161cb90155fb183eed8355f4 from pypy

cfbolz:
https://bitbucket.org/pypy/pypy/commits/7f5d1ceedd1dffa5161cb90155fb183eed8355f4

The eval was not just removing the quotes, but also turning `\n` into linefeed characters.

### DIFF
--- a/lib/eco/pgviewer.py
+++ b/lib/eco/pgviewer.py
@@ -172,10 +172,11 @@ class PGViewer(object):
         return dotnode
 
 def splitline(line, re_word = re.compile(r'[^\s"]\S*|["]["]|["].*?[^\\]["]')):
+    import ast
     result = []
     for word in re_word.findall(line):
         if word.startswith("\""):
             assert word[-1] == "\""
-            word = word[1:-1]
+            word = ast.literal_eval(word)
         result.append(word)
     return result


ltratt:
Aha, that I didn't know. Good catch!

Can we stick the `import` at the top of the file with all the others, for consistency?

ptersilie:
To be honest. It doesn't really matter as line-feeds are handled completely different by Eco anyways. So the current fix works good enough without the need for an import so I'd probably just leave it at that.

cfbolz:
It's not just for line-feeds but also other escaped characters.

ltratt:
Maybe I can ask a dumb question: clearly the `eval` could do a lot of things. What use-cases are there for it doing the complex stuff in this particular instance?

cfbolz:
Well, it's obvious that a full eval is a bad idea. But using `literal_eval` supports just a small subset, which is evaling python literals. If you prefer that, it seems it's also possible to write this kind of code: `word[1:-1].decode('string_escape')`.

ltratt:
Let me ask a dumb question: do we _need_ escape characters to be converted in this instance? [I must admit, when I looked at this yesterday, I couldn't see a use-case for this, but maybe I missed it.]

cfbolz:
In PyPy it's clearly needed. In Eco, I have no idea (one of the risks of copy-pasting huge amounts of code whole-sale).

ptersilie:
Sorry, didn't see you already made a PR (thought this was an issue). In that case I don't mind merging it. It certainly catches more cases than the current implementation even if we probably never run into those cases.

cfbolz:
But if it's not needed in Eco, then the whole condition `if word.startswith("\""):` can be killed.

ltratt:
Personally, then, I'm fine with this PR provided we move the import to the top.

ptersilie:
Sounds good.

ltratt:
Is this PR still relevant? [I think it is, but it's languished for a long while!]

ltratt:
@ptersilie This PR is old, but still mergeable. Should it be merged?

ptersilie:
Merged. Finally.

ltratt:
I think this PR matured nicely ;)

Issue #151, opened 2016-07-08
Pop up a dialog box if the user tries to execute a syntactically incorrect program


vext01:
Minor quibble. Just because it's not always obvious why a program has no output when executed.

ltratt:
+1

I sometimes get a bit confused by this, until I remember to look at the terminal.

I think this would be a good one for @djmcalleja to have a quick look at.

djmcalleja:
Sure, I can do that.

ptersilie:
:+1: Something like `Failed to export program. Please check if your program is syntactically correct before exporting`.

vext01:
Why "please check", eco should know if it's correct or not, no?

ptersilie:
maybe "please check that ... is valid before exporting"

ptersilie:
I dunno. Was just a suggestion. Could also say "Sorry dude. Can't export cus invalid."

vext01:
How about:

```
Eco cannot execute this program because it is syntactically incorrect
```

Replace "execute" with "export" as context requires.

vext01:
(But I do like your slang error message too)

ltratt:
I prefer "syntactically invalid", but I think Edd's message is otherwise about right.

ptersilie:
Okay let's go with `Eco cannot execute/export this program because it is syntactically invalid`.

djmcalleja:
I'm supposed to make a pull request for this, right? Who should I assign it to?

snim2:
For issues, usually the person who raised the original issue (so, @vext01 )

djmcalleja:
Right, ok

PR #152, opened 2016-07-08
Message appears when trying to execute/export program with syntax errors

djmcalleja:
Fixes https://github.com/softdevteam/eco/issues/151

### DIFF
--- a/lib/eco/debug.py
+++ b/lib/eco/debug.py
@@ -13,22 +13,23 @@ class Debugger(QObject):
         # incoming signals
         self.signal_command = SIGNAL("command")
         self.signal_start = SIGNAL("start")
-        # outcoming signals
+        # outgoing signals
         self.signal_done = SIGNAL("finished")
         self.signal_output = SIGNAL("output")
         self.signal_toggle_buttons = SIGNAL("disablebuttons")
+        self.signal_execute_fail = SIGNAL("executefail")
         self.telnet = None
 
     def start_pdb(self):   
         self.breakpoints = {'keep': [], 'del': []}      
         self.line_read_until = '(Pdb)'   
         self.process = self.window.getEditor().tm.export(debug=True)
-        flags = fcntl(self.process.stdout, F_GETFL)
-        fcntl(self.process.stdout, F_SETFL, flags | O_NONBLOCK)       
         if not self.process:
-            self.emit(self.signal_output, "Cannot run debugger process, cannot debug.")
+            self.emit(self.signal_execute_fail)
             self.exit()
             return
+        flags = fcntl(self.process.stdout, F_GETFL)
+        fcntl(self.process.stdout, F_SETFL, flags | O_NONBLOCK)
         self.telnet = None
         attempt = 0
         while not self.telnet:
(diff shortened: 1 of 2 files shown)


snim2:
Great! BTW if you say "Fixes Issue ..." then GitHub will automatically close the issue when this has been merged.

lib/eco/debug.py
```diff
@@ -13,22 +13,23 @@ def __init__(self, window):
...
         self.line_read_until = '(Pdb)'   
         self.process = self.window.getEditor().tm.export(debug=True)
-        flags = fcntl(self.process.stdout, F_GETFL)
-        fcntl(self.process.stdout, F_SETFL, flags | O_NONBLOCK)       
         if not self.process:
-            self.emit(self.signal_output, "Cannot run debugger process, cannot debug.")
+            self.emit(self.signal_execute_fail)
             self.exit()
             return
+        flags = fcntl(self.process.stdout, F_GETFL)
```
vext01, line 31: I've never spent much time on eco, so here's some dumb questions:

Why does popping up an error require changes to file descriptor flags?

lib/eco/eco.py
```diff
@@ -891,7 +892,9 @@ def cli_export(self, source, dest, fast):
             self.tm.fast_export(language_boxes, dest, source=source)
         else:
             self.tm.load_file(language_boxes)
-            self.tm.export(dest, source=source)
+            ex=self.tm.export(dest, source=source)
```
vext01: you need spaces around that assignment.

lib/eco/eco.py
```diff
@@ -1174,7 +1179,12 @@ def exportAs(self):
         if path:
             self.save_last_dir(str(path))
             ed.export_path = path
-            self.getEditor().tm.export(path, source=ed.filename)
+            ex=self.getEditor().tm.export(path, source=ed.filename)
```
vext01: and here

vext01:
@ptersilie a lot of this is QT event loop stuff, does it look OK to you?

ltratt:
@vext01 For these sorts of PRs, I think you have to run them locally to check.

vext01:
I mean in terms of code quality. I am going to try it out though.

lib/eco/debug.py
djmcalleja, line 31: The line hasn't changed but I had to move it to after line 28, which checks if `self.process` has a value

ltratt, line 31: Note that this is a standard fcntl idiom -- you need to read the flags before setting it (in the next line) to non-blocking (maintaining other flags).

vext01:
FWIW, this works from a functional standpoint.

lib/eco/debug.py
vext01, line 31: Yes, I know fcntl ;)

I think this is OK.

vext01:
Looks OK from where I am standing, but I think you should ask Lukas, as I don't really know what I'm doing when it comes to QT.

ltratt:
If it runs OK, and the patch doesn't look insane, I think we're OK :)

vext01:
Alright then :P

ltratt:
Um, this stuff still needed squashing. Too late now though...

vext01:
Yeah, I accidentally hit merge. Sorry.

ptersilie:
@vext01 I think for most PRs you can squash during merging on Github itself.

PR #153, opened 2016-07-08
Fix export bug

djmcalleja:
Found a small bug where the syntax error dialog box from PR https://github.com/softdevteam/eco/pull/152 would appear after a successful export. This fixes that. Running programs straight from Eco worked fine.

### DIFF
--- a/lib/eco/eco.py
+++ b/lib/eco/eco.py
@@ -470,7 +470,7 @@ class PreviewDialog(QtGui.QDialog):
             self.ui.textEdit.setText(text)
         elif index == "Default":
             text = self.tm.export("/dev/null")
-            if text is None or text is False:
+            if not text:
                 text = ""
                 self.show_export_fail_message()
             self.ui.textEdit.setText(text)
@@ -893,7 +893,7 @@ class Window(QtGui.QMainWindow):
         else:
             self.tm.load_file(language_boxes)
             ex = self.tm.export(dest, source=source)
-            if not ex:
+            if ex == False:
                 self.show_export_fail_message()
         QApplication.quit()
         sys.exit(1)
@@ -1168,7 +1168,7 @@ class Window(QtGui.QMainWindow):
             ed.export_path = QFileDialog.getSaveFileName()
         if ed.export_path:
             ex = self.getEditor().tm.export(ed.export_path, source=ed.filename)
-        if not ex:
+        if ex == False:
             self.show_export_fail_message()
 
     def exportAs(self):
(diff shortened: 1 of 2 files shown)


ltratt:
Might it be clearer to do something like `assert x is not None` rather than `is x == False`?

djmcalleja:
Oops, my last comment wasn't clear. I meant, the condition should be true only if `ex` is False. If it's anything else then ignore it.

lib/eco/eco.py
```diff
@@ -470,7 +470,7 @@ def change(self, index):
             self.ui.textEdit.setText(text)
         elif index == "Default":
             text = self.tm.export("/dev/null")
-            if text is None or text is False:
+            if text is None or text == False:
```
snim2: `if not text` should catch both of these cases:

``` python
>>> not None
True
>>> not False
True
>>> 
```

lib/eco/treemanager.py
```diff
@@ -1567,16 +1567,18 @@ def export(self, path=None, run=False, profile=False, source=None, debug=False):
...
 
         if str(path).endswith(".aterms"):
             return self.export_aterms(path)
 
         lang = self.parsers[0][2]
         if lang == "Python + Prolog":
             self.export_unipycation(path)
+            return True
         elif lang == "HTML + Python + SQL":
             self.export_html_python_sql(path)
```
snim2, line 1580: Does `self.export_html_python_sql(path)` not return a value?

djmcalleja, line 1580: It just writes to the file without returning anything.

snim2, line 1580: Well, this is a bit baffling to me, but I still get an error message if the HTML+Python+SQL file is syntactically incorrect.

snim2:
Aside form the `if` statements this looks fine to me, and it does work. You will need to squash these changes, so before you do, please make sure that trailing whitespace has been deleted in both the files you have touched here.

lib/eco/treemanager.py
djmcalleja, line 1580: Do you mean it still shows the message if the syntax is correct? If the message appears when trying to export a file with incorrect syntax then it's working properly.

snim2, line 1580: No I mean it is working correctly, but it wasn't quite clear to me why, but I guess your `text` is `None` in that case?

djmcalleja, line 1580: It works because the export function returns False right at the top if the syntax is incorrect, this is the same for both exporting and running in Eco.

snim2, line 1580: Ah, thanks, I missed that completely.

snim2:
Thanks, I'll merge when Travis has finished

djmcalleja:
Ok thanks

PR #155, opened 2016-07-15
Fixed string regex for eco grammars

### LINKED ISSUE #154: Parsing not triggering?
1) Load in `basiccalc.eco`
2) Add the following text on a new line after `%implicit_ws=true`: `%left "` (i.e. don't close the `"`). Note that the parsing status is green when I think it should be red.
3) Select the whole file, cut it (Ctrl-X), then paste it back in. The parsing status is now red.

ptersilie:
The old regex allowed strings that consisted of a single quote. This has been fixed by changing [^"] to [^"\r] in the `terminal` definition.

Fixes #154


ltratt:
Thanks!

Issue #121, opened 2016-04-28
Overlay does not take account of language boxes of different heights


snim2:
The overlay currently assumes that every line in the current document has the same height. This isn't always true, and it certainly isn't true if some of the language boxes contain images.

snim2:
This is still a valid issue, but it isn't going to be addressed.

PR #156, opened 2016-07-18
JRuby HUD. Includes callgraph, eval() strings and argument types.

snim2:
This PR contains the eval() and types information from the recent JRuby callgraph changes. This was trickier than expected, because we don't yet have a generic solution to Issue #121 

The solution here allows the user to choose which visualisation they want to show in a heads-up display, out of a choice of:
- No HUD
- Callgraph
- eval() strings
- argument types
- heat map (currently heat map information is not available from JRuby)

Examples:
## HUD off

hud-off
## Callgraph

hud-callgraph
## eval() strings

hud-eval
## argument types

hud-types

### DIFF
--- a/lib/eco/export/jruby.py
+++ b/lib/eco/export/jruby.py
@@ -25,7 +25,7 @@ import os.path
 import tempfile
 import subprocess
 
-from incparser.annotation import Annotation, ToolTip, Railroad
+from incparser.annotation import Annotation, Footnote, ToolTip, Railroad, Eval, Types
 
 from incparser.astree import EOS
 from grammar_parser.gparser import MagicTerminal, IndentationTerminal
@@ -33,13 +33,18 @@ from grammar_parser.gparser import MagicTerminal, IndentationTerminal
 from PyQt4.QtCore import QSettings
 
 
-class JRubyMorphismLine(Annotation):
+class JRubyCallGraph(Annotation):
     """Annotation for JRuby callgraph railroad diagrams.
     """
 
     def __init__(self, annotation):
         self._hints = [Railroad()]
-        super(JRubyMorphismLine, self).__init__(annotation)
+        super(JRubyCallGraph, self).__init__(annotation)
+
+    def has_hint(self, klass):
+        if klass == Railroad:
+            return True
+        return False
 
     def get_hints(self):
         return self._hints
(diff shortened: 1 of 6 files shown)


lib/eco/export/jruby.py
```diff
@@ -435,7 +508,26 @@ def annotate_tree(self, src_file_name, log_file_name):
                                         def_filename,
                                         method.name,
                                         { method.name : method.is_mega },
-                                        JRubyMorphismLine)
+                                        JRubyCallGraph)
+                    # Eval strings interpreted at runtime.
+                    if def_eval_strings:
+                        self._annotate_text(def_lineno,
```
ltratt: These one-parameter-per-line function calls make reading code really difficult, as so little fits on a screen. Let's please reformat all of them so that there are multiple parameters per line. I tend to wrap to 100 chars per line, but there's also a good argument for 80 chars. Even with the latter, this can be reformatted to:

``` python
                        self._annotate_text(def_lineno, def_filename,
                                            method.name,
                                            ("%s" % ", ".join(def_eval_strings)),
                                            JRubyEvalStrings)
```

lib/eco/incparser/annotation.py
```diff
@@ -43,6 +43,18 @@ class ToolTip(Hint):
     pass
 
 
+class Eval(Hint):
```
ltratt, line 46: Will these classes (Eval and Types) have some concrete data in the future, or will they only ever be placeholders?

snim2, line 46: The hints are only placeholders (they are "hints" for the bits of Eco that do the visualisation, so that the rendering code knows which annotations to use and which to ignore). I expect this sort of thing could be done better, but this fits with the model of the code we already have, a large-ish refactoring is probably worth doing soon.

ltratt, line 46: OK, sounds sensible.

ltratt:
Modulo the function call reformatting, this looks good, although it will need rebasing against master before it can be merged.

snim2:
OK, this has been squashed and brought up to date.

ltratt:
OK, thanks.

[BTW, next time, let's leave a separate commit that only fixes trailing whitespace. It makes it easier for dumbies like me to follow the diffs.]

snim2:
Sorry about that, my editor is set up to remove trailing whitespace automatically, so I never see this stuff until I've already committed it. Next time I'll `commit -p`

ltratt:
Understood, though I'd strongly advise turning that feature off, otherwise (from bitter experience), we'll keep getting tripped up by it.

PR #157, opened 2016-07-27
Put callgraph HUD into its own dock

ptersilie:
Allow to hide the callgraph HUD without having to hide the parsing status as well.


snim2:
That looks nice!

PR #161, opened 2016-09-01
Fix the HUD to work with languages other than JRuby

### LINKED ISSUE #107: FIX heatmap visualisation
Some recent commits have broken the heatmap visualisation. ~~This is partly because some external paths an binaries are set via the Settings dialog and some are not.~~

snim2:
- Fixes an old bug in cPython profiling
- Greys out HUD buttons for languages that cannot provide HUD information 
  [see this video for an example](https://youtu.be/zFFjqm90zDc)
- Displays onle the tooltips etc. that the user has requested via the HUD. 

e.g. user has selected callgraph information:

hud-with-tooltip

Here the user has selected "No HUD" so the tooltip is not displayed:

no-hud-no-tooltip

Fixes #107

### DIFF
--- a/lib/eco/eco.py
+++ b/lib/eco/eco.py
@@ -656,11 +656,15 @@ class Window(QtGui.QMainWindow):
 
         # Set up HUD radio buttons.
         self.ui.hud_off_button.setChecked(True)
+        self.connect(self.ui.hud_off_button, SIGNAL("clicked()"), self.hud_off_toggle)
         self.connect(self.ui.hud_callgraph_button, SIGNAL("clicked()"), self.hud_callgraph_toggle)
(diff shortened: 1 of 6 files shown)


ltratt:
Maybe we should do a PR that's just whitespace removal? I must admit, it is a bit distracting to have it in every PR. [I agree, though, that getting rid of it is generally a good thing.]

snim2:
Hopefully this is the last of it. Do you want me to cherry-pick that commit into another PR?

ltratt:
Ah, if it's the last one, don't worry about it.

snim2:
This is ready for review now.

ltratt:
Looks good. Do you want to squash and then I can merge?

snim2:
OK, squashed!

ltratt:
Will merge once tests have completed.

snim2:
Thanks

PR #162, opened 2016-09-02
Speed up travis-ci.

snim2:


snim2:
@ptersilie so, that Travis run skipped the slow tests. If you choose to merge this, the next Travis run should be on master (and not a PR) and so should run slow.

PR #167, opened 2016-09-19
Solving multiline comments and strings with MultiTextNodes

ptersilie:
IN DEVELOPMENT - NOT READY TO MERGE

This PR solves our long ongoing problem of comments and strings containing newlines or language boxes, which where especially hard to deal with in combination with the indentation stuff in the python parser.

The previous solution solved multiline comments/strings on the parser level. This new solution solves it on the lexer level, by putting all elements of a multiline structure into a `MultiTextNode`.

So for instance

``` python
x = "This is a multiline
string"
```

will create a `MutliTextNode` like this:

``` python
MultiTextNode([TextNode("\"This is a multline"), TextNode("\n"), TextNode("string\"")])
```

The same works for language boxes, e.g.

``` python
x = "Here is a language box |SELECT * FROM table|! Nice!"
```

becomes

``` python
MultiTextNode([TextNode("\"Here is a language box"), TextNode(LBox("<SQL>")), TextNode("! Nice!\"")])
```

These changes allow us to move the indentation algorithms back into a separate phase between lexing and parsing instead of having to generate indetation tokens _during_ the parsing phase (which becomes especially complicated with history and error recovery).


ptersilie:
Will be merged together with error recovery.

PR #171, opened 2017-01-23
Error recovery, node reuse, multinodes

ptersilie:
This PR will merge multiple branches I've been working on into master:

Error recovery: After finding an error the parser can continue to parse the rest of the program and merge successfully parsed subtrees into the current parse tree. Includes out-of-context-analysis which can parse subtrees after the error node within the isolated subtree, and (currently disabled) retainability which integrates subtrees within the iso tree before the error node.

Node reuse: When reducing nodes into a new nonterminal this feature enables the reuse of the former parent of one of the nodes. This avoids recreating nonterminal nodes that are already in the parse tree.

Multinodes: Enables the lexer to create Terminal nodes that consists of multiple other terminals. This is needed to implement language boxes and newlines within strings and comments.


ptersilie:
I have a few questions on how to approach this PR.

1) There are currently some tests failing. At the moment I expect them to fail, but they will have to fixed in the future. One tests retainability which is currently disabled as there is a bug in the retainability checks which can lead to loops within the parse tree. The other tests a bug in the error recovery algorithm which leads to a valid parse tree, but the parse tree is not correct. I already have an idea on how to fix the latter, but I probably do that in a separate PR after we've merged this one. Shall I just merge this with the failing tests?

2) This PR contains a lot of commits as it contains 3 different branches that have already been merged together. I tried squashing this, but it has become to difficult to rearrange the commits in a meaningful way. My plan was to just squash this down into a single commit into another branch and merge that branch. This way I won't loose the commit history, which I'd like to keep for now.

ltratt:
I'm happy to have tests that don't pass, but can we mark them as `xfail` (or whatever it is) so that they don't cause Travis to barf?

I don't understand the second point though: it seems to be saying "I'll squash them down into a single commit so that I maintain commit history"?!

Oh, and before we do a review, you should probably rebase against master and fix the conflicts that will prevent a merge.

ptersilie:
Ah, there is a marker for tests like that. Good. I'll try that.

On the second point. My plan was to squash everything into a different branch which I then merge and keep this branch as it is (if something like this is possible). I don't have a better idea unfortunately, but I'm open to suggestions.

ltratt:
I don't understand what keeping the branch as-is would achieve, but I may be missing out on something obvious.

ptersilie:
Merged with master and moved to #173.

PR #175, opened 2017-02-06
Fix keyword priorities in Java lexer

### LINKED ISSUE #172: Java lexer slightly broken?
The current Java 1.5 grammar has this part in the lexer:

```
NULL_LITERAL:"null"
IDENTIFIER:"[a-zA-Z_][a-zA-Z0-9_]*"

CONST:"const"
GOTO:"goto"
STRICTFP:"strictfp"
ELLIPSIS:"ellipsis"
ENUM:"enum"
```

In other words, we match `IDENTIFIER` before other things that look like identifiers. That means that code like this:

```java
class C {
    public void m() {
        int enum = 2;
    }
}
```

is treated as correct, even though javac rejects it.

I think the fix is simply to reorder things a bit and make sure that `IDENTIFIER` is the last thing in the file:

```
CONST:"const"
GOTO:"goto"
STRICTFP:"strictfp"
ELLIPSIS:"ellipsis"
ENUM:"enum"

NULL_LITERAL:"null"
IDENTIFIER:"[a-zA-Z_][a-zA-Z0-9_]*"
```

ptersilie:
Fixes #172


ltratt:
Yep, looks good to me. [I checked it locally, since GitHub doesn't yet diff `.eco` files ;)]

ltratt:
[Sorry, the tests still take a long time to run before I can merge!]

PR #176, opened 2017-02-06
Reuse old indentation nodes

ptersilie:
This is a small optimisation that tries to reuse more indentation nodes than the previous approach. Until now indentation nodes were only reused if all indentation nodes in a line stayed the same. Otherwise the algorithm would first remove all indentation nodes and then insert the new ones. This sometimes lead to indentation nodes being replaced by the same node.

The new algorithm looks at each indentation node individually and compares them with the new indentation node that replaces it. Equal nodes are then skipped and not removed and re-inserted.

### DIFF
--- a/lib/eco/indentmanager.py
+++ b/lib/eco/indentmanager.py
@@ -165,14 +165,7 @@ class IndentationManager:
         """Insert generated indentation tokens into the token stream if they
         differ from current tokens"""
         # test if indent nodes have changed
-        changed = self.indentation_nodes_changed(bol, new)
-        if changed:
-            self.changed = True
-            # remove old indentation nodes
-            self.remove_indentation_nodes(bol)
-            # and insert new ones
-            for node in new:
-                bol.insert_after(node)
+        self.merge_nodes(bol, new)
 
         # generate last lines dedent
         if self.next_line(bol) is None: # this is the last line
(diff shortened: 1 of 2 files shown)


lib/eco/indentmanager.py
```diff
@@ -196,32 +189,38 @@ def apply_nodes(self, new, bol):
...
 
-        previous_nodes.reverse()
-        if len(previous_nodes) != len(nodes):
-            return True
-        for i in range(len(nodes)):
-            if nodes[i].symbol != previous_nodes[i].symbol:
-                return True
-        return False
+        for i in range(len(previous_nodes)):
+            try:
```
ltratt: Wouldn't this be tidier as `if len(newnodes) == 0: XXX else: YYY`?

lib/eco/indentmanager.py
```diff
@@ -196,32 +189,38 @@ def apply_nodes(self, new, bol):
...
+                # remove old, insert new
+                oldnode.replace(newnode)
+                last_node = newnode
+            else:
+                if oldnode.deleted:
+                    oldnode.deleted = False
+                last_node = oldnode
+
+        for n in newnodes:
+            # insert remaining newnodes
```
ltratt, line 222: I'm not sure this comment adds anything useful.

ptersilie:
I just had an idea where we potentially could reuse nodes that have previously been deleted, by undeleting them. Let's assume we have the program

```
def x():
    p
```

and then change the program to

```
def x():
p
```

and then back to the original again.

When the indentation algorithm tries to insert the new indentations, the in version 2 deleted indentation nodes still exist in the parse tree, marked as `deleted`, since the parse tree hasn't been successfully reparsed (only then are `deleted` nodes removed from the tree). The idea is to check if these `deleted` nodes can be reused and then just undelete them instead of inserting the same indentation node again.

I will have to run some tests to see if this doesn't mess anything up with the versioning.

ltratt:
Do you mean if someone does undo then redo?

lib/eco/indentmanager.py
ptersilie: It's kinda part of `remove leftover previous nodes` and `remove old, insert new`. I don't think it adds less than those comments.

ltratt: As you wish. It seems on the obvious side to me, but this is partly a matter of personal choice.

lib/eco/indentmanager.py
ptersilie: You mean

```
if len(newnodes) == 0:
    newnode = newnodes.pop()
else:
    previous_nodes[i].remove()
    continue
```

Think I just chose `try` as it's less code to write than the `if` statement. ;)

ltratt: Yes, the `if` version is much clearer and more obvious (not to mention faster!).

ptersilie:
No, I mean someone deletes the space and then manually types it back in. Redo/Undo wouldn't regenerate the indentation nodes at all, and also does not change the tree, nor reparses it.

ltratt:
Can I suggest that we leave that optimisation to another pull request, after everything else has been gotten working? Just in case it turns into an effort tar pit!

ptersilie:
Yes I don't mind doing that.

ptersilie:
Although the change will probably just be a single line if I'm not wrong.

ltratt:
Up to you!

ptersilie:
I'll do some quick tests and let you know by tomorrow.

ltratt:
OK.

ptersilie:
Okay, this optimisation is a bit more complicated. So let's just merge this first.

ltratt:
I think we just need to fix the `if` comment, then we can merge.

ptersilie:
Okay addressed this now. If you are happy to merge, you can just squash this down into a single commit using Github's "Squash and merge".

ltratt:
I think if I do "squash and merge" then I don't get a merge commit (which we want)?

ptersilie:
Fair enough. Squashed manually now.

ltratt:
Thanks... although I still see two commits?

ptersilie:
Me too. But I definitely squashed and pushed. Interesting.

ltratt:
Maybe just do the force push again?

ptersilie:
It tells me everything up to date. And my gitk shows me only one commit.

ltratt:
They've arrived!

ptersilie:
Ah as I thought. Github just being a bit slow.

PR #180, opened 2017-02-14
Avoid error highlighting wrong nodes

ptersilie:
Fixes the highlighting of nodes that are equal to the error node


ltratt:
In general, it's not safe to use `id` to store references to things, as their ID could change. It's probably simpler to use a linear scan and, when needed, use the `is` operator.

ltratt:
Looks good to me. Squash and then I'll merge?

ptersilie:
Squashed!

PR #185, opened 2017-02-24
Improved error recovery and detection

ptersilie:
Before this PR isolated subtrees weren't rechecked on subsequent parses, allowing the parser to accept incorrect subtrees. The problem has been fixed and the corresponding test reactivated.

This PR also improves the detection of errors that are hidden within previously isolated (and thus in subsequent parses, skipped) subtrees.

Together these changes allow the editor to show multiple errors in different parts of the program and also allows fixing or introducing new errors within already isolated subtrees.

### DIFF
--- a/lib/eco/incparser/error_recovery.py
+++ b/lib/eco/incparser/error_recovery.py
@@ -75,6 +75,14 @@ class RecoveryManager(object):
 
         self.rejects.add(node)
 
+        # There is no point in isolating nodes that don't have changes, even if
+        # they are valid isolation trees. Ultimately the parse will fail again
+        # as no changes have been isolated, and a bigger isolation area will be
+        # picked. This is a deviation from Wagner's algorithm, which doesn't seem
+        # to check for this.
+        if not node.has_changes(self.previous_version):
+            return False
+
         if len(node.children) == 0:
             return False
 
--- a/lib/eco/incparser/incparser.py
+++ b/lib/eco/incparser/incparser.py
@@ -216,8 +216,7 @@ class IncParser(object):
                         la = result
 
             else: # Nonterminal
-                #iso_and_changed = la.isolated and self.surrounding_context_changed(la)
-                if la.has_changes() or needs_reparse or la.has_errors():# or iso_and_changed:
+                if la.has_changes() or needs_reparse or la.has_errors() or self.iso_context_changed(la):
                     #la.changed = False # as all nonterminals that have changed are being rebuild, there is no need to change this flag (this also solves problems with comments)
                     la = self.left_breakdown(la)
                 else:
@@ -303,6 +302,7 @@ class IncParser(object):
         elif isinstance(element, Shift):
             self.validating = False
             self.shift(la, element)
+            la.local_error = la.nested_errors = False
             return self.pop_lookahead(la)
 
         elif isinstance(element, Reduce):
@@ -554,9 +554,12 @@ class IncParser(object):
         assert goto != None
 
         # save childrens parents state
+        has_errors = False
         for c in children:
-            c.local_error = False
-            c.nested_errors = False
+            if c.has_errors():
+                has_errors = True
+            if c.isolated:
+                has_errors = True
             if not c.new:
                 # just marking changed is not enough. If we encounter an error
                 # during reduction the path from the root down to this node is
@@ -570,6 +573,7 @@ class IncParser(object):
             new_node.changed = False
             new_node.deleted = False
             new_node.isolated = False
+            new_node.nested_errors = has_errors
             new_node.set_children(children)
             new_node.state = goto.action # XXX need to save state using hisotry service
             new_node.mark_changed()
@@ -778,23 +782,14 @@ class IncParser(object):
                 return c
         return None
 
-    def surrounding_context_changed(self, node):
-        # find isolation trees last terminal
-        while isinstance(node, Nonterminal):
-            if node.children:
-                node = node.children[-1]
-                continue
-            else:
-                # find left sibling
-                while not node.left:
-                    node = node.parent
-                continue
-        # found terminal
-        if node.next_term.changed:
-            return True
-
-        if node.next_term is not node.get_attr("next_term", self.prev_version):
-            return True
+    def iso_context_changed(self, node):
+        # Currently catches more cases than neccessary. Could be made more
+        # accurate by finding the next terminal reachable from node (including
+        # deleted ones)
+        if not node.isolated:
+            return False
+        la = self.pop_lookahead(node)
+        return la.has_changes()
 
     def next_terminal(self, node):
         n = self.pop_lookahead(node)
--- a/lib/eco/incparser/incparser.py
+++ b/lib/eco/incparser/incparser.py
@@ -141,11 +141,12 @@ class IncParser(object):
         self.inc_parse([], True)
 
     def inc_parse(self, line_indents=[], needs_reparse=False, state=0, stack = []):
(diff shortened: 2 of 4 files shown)


lib/eco/incparser/error_recovery.py
```diff
@@ -36,6 +36,10 @@ def recover(self, error_node):
         # Get the character offset of the top node on the stack
         error_offset = self.stack_offset(len(self.stack))
 
+        # XXX When searching for an isolation node and we discover an iso node on
```
ltratt: Not sure I understand this. If you're searching for an isolation node, discovering an iso (short for "isolation"?) node is what you want, right?

lib/eco/incparser/error_recovery.py
```diff
@@ -71,10 +75,19 @@ def recover(self, error_node):
     def is_valid_iso_subtree(self, node, error_offset):
         if node in self.rejects:
             # Node has already been rejected as a candidate for recovery
+            # XXX node.isolated = False # reset isolations that didn't work?
```
ltratt: Commented out code.

lib/eco/incparser/error_recovery.py
```diff
@@ -71,10 +75,19 @@ def recover(self, error_node):
...
             return False
 
         self.rejects.add(node)
 
+        # There is no point in isolating nodes that don't have changes, even if
+        # they are valid isolation trees. Ultimately the parse will fail again
+        # as no changes have been isolated, and a bigger isolation area will be
+        # picked. This is a derivation on Wagner's algorithm, who doesn't seem
```
ltratt: "derivation on" -> "deviation from"?

lib/eco/incparser/incparser.py
```diff
@@ -216,8 +227,7 @@ def inc_parse(self, line_indents=[], needs_reparse=False, state=0, stack = []):
                         la = result
 
             else: # Nonterminal
-                #iso_and_changed = la.isolated and self.surrounding_context_changed(la)
-                if la.has_changes() or needs_reparse or la.has_errors():# or iso_and_changed:
+                if la.has_changes() or needs_reparse or la.has_errors() or self.iso_context_changed(la):
                     #la.changed = False # as all nonterminals that have changed are being rebuild, there is no need to change this flag (this also solves problems with comments)
```
ltratt: Commented out code.

lib/eco/incparser/incparser.py
```diff
@@ -385,6 +400,9 @@ def pass2(self, node, offset, error_offset):
             offset += c.textlength()
 
     def is_retainable_subtree(self, node):
+        # Subtrees are currently always unretainable due to a potential bug in
+        # Wagners algorithm
+        return False
         if node.new:
```
ltratt: Unreachable code.

lib/eco/incparser/incparser.py
```diff
@@ -385,6 +400,9 @@ def pass2(self, node, offset, error_offset):
             offset += c.textlength()
 
     def is_retainable_subtree(self, node):
+        # Subtrees are currently always unretainable due to a potential bug in
+        # Wagners algorithm
```
ltratt: Can you maybe briefly explain what the potential bug is?

lib/eco/incparser/incparser.py
```diff
@@ -496,22 +516,27 @@ def out_of_context_analysis(self, node):
               logging.debug("OOC analysis of %s failed. Error on %s.", node, temp_parser.error_nodes)
               self.error_nodes.extend(temp_parser.error_nodes)
               node.log[("left", self.prev_version)] = saved_left
-              self.isolate(node) # XXX actually need to call refine here
-              node.isolated = True
+              temp_error = node.isolated
+              # XXX has this subtree already been isolated/refined within
```
ltratt: Does this mean "we don't have a way of telling if this subtree has been isolated/refined, so we have to do an expensive check" or ...?

lib/eco/incparser/incparser.py
```diff
@@ -554,9 +580,12 @@ def reduce(self, element):
         assert goto != None
 
         # save childrens parents state
+        has_errors = False
         for c in children:
-            c.local_error = False
-            c.nested_errors = False
+            if c.has_errors():
```
ltratt: `if c.has_errors() or c.isolated` (and then you can remove the `if` beneath)?

ltratt:
OK, a few minor comments, nothing too complicated I think.

lib/eco/incparser/error_recovery.py
ptersilie: I meant an already isolated node (i.e. the algorithm chose an isolated tree that afterwards turns out to not be a big enough isolation so it needs to find a bigger isolation; in which case we need to unmark the previous isolation tree again). I'm not sure yet if this can actually happen. It happened once but only due to a bug so it disappeared after I fixed the bug. But I'm not 100% certain that there isn't a rare scenario that can cause this again.

ltratt: Can we add an `assert` that catches the bad case, should it ever manifest again?

lib/eco/incparser/error_recovery.py
ptersilie: This is part of the above comment and might be the fix for the problem described above (if it is a problem that is). Maybe I should make that more clear.

lib/eco/incparser/error_recovery.py
ptersilie: Thanks, will fix.

lib/eco/incparser/incparser.py
ptersilie: Will remove.

lib/eco/incparser/incparser.py
ptersilie: I can try. It's quite difficult to explain. I do have a test for it. I'm gonna try and add a more detailed explanation to the test and the reference the test here.

lib/eco/incparser/incparser.py
ptersilie: Yes. I just don't want to delete that code because the implementation is correct. So once I found out what the problem is with the retainability algorithm is I just have to remove the `return False`.

ltratt: OK, I think we need to make that clearer.

lib/eco/incparser/incparser.py
ptersilie: This is again a comment for me to think a bit more about the algorithm. In theory the out-of-context analysis should try and parse the subtree and if unsuccessful isolated it. So when we reach this point it *should* already be isolated. I'll have to double check that this is true because then we can remove the `self.isolate()` call.

lib/eco/incparser/incparser.py
ptersilie: Will fix.

lib/eco/incparser/error_recovery.py
ptersilie: Good idea.

ptersilie:
Okay, hopefully addressed all of your comments.

ltratt:
OK, this looks good. Do you want to do some squashing before I merge?

ptersilie:
Squashed!

ptersilie:
Just found a bug. Don't merge yet!

ltratt:
OK.

ptersilie:
Okay, bug is fixed. Think we can merge this now.

ltratt:
Do you want to squash or should I merge as-is?

ptersilie:
I think this is fine as is.

PR #186, opened 2017-03-01
Show node changes that are actually responsible for an error

ptersilie:
When a syntax error happens the reason for that error is rarely the node that causes the parsing error. Mostly another node had been changed earlier and the parser only fails later trying to make sense of the change. This PR also marks all changes that have been made between the last correct program to the current erroneous program and highlights them in the editor (alongside the syntax error).

### DIFF
--- a/lib/eco/treemanager.py
+++ b/lib/eco/treemanager.py
@@ -370,6 +370,7 @@ class TreeManager(object):
         self.changed = False
         self.last_search = ""
         self.version = self.global_version = 1
+        self.reference_version = 0
         TreeManager.version = 1
         self.last_saved_version = 1
         self.savenextparse = False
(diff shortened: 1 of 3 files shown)


lib/eco/treemanager.py
```diff
@@ -521,6 +522,23 @@ def get_error(self, node):
...
+                    if n is node:
+                        return True
+        return False
+
+    def get_error_presentation(self, node):
+        for p in self.parsers:
+            if p[0]:
+                for n, pres in p[0].error_pres:
+                      if n is node:
+                          #prev_name = node.get_attr("symbol.name", self.reference_version)
```
ltratt: Commented code.

ptersilie: Ah damn. I thought I deleted that.

ltratt:
I think I can merge after that comment is dealt with.

ptersilie:
Fixed. If OK I'll squash and you can merge.

ltratt:
Yep, please squash away!

ptersilie:
Done!

PR #187, opened 2017-03-06
Import bug

ptersilie:
Some fixes related to saving and loading Eco files


ltratt:
If I load `range.eco`, it does load but as soon as I try and edit a character I get:
```python
  File "/home/ltratt/src/eco/lib/eco/nodeeditor.py", line 857, in keyPressEvent
    self.tm.key_normal(text)
  File "/home/ltratt/src/eco/lib/eco/treemanager.py", line 1041, in key_normal
    self.reparse(node, need_reparse)
  File "/home/ltratt/src/eco/lib/eco/treemanager.py", line 1865, in reparse
    parser.inc_parse()
  File "/home/ltratt/src/eco/lib/eco/incparser/incparser.py", line 259, in inc_parse
    self.reduce(element)
  File "/home/ltratt/src/eco/lib/eco/incparser/incparser.py", line 628, in reduce
    self.interpret_annotation(new_node, element.action)
  File "/home/ltratt/src/eco/lib/eco/incparser/incparser.py", line 647, in interpret_annotation
    astnode = annotation.interpret(node)
  File "/home/ltratt/src/eco/lib/eco/grammar_parser/bootstrap.py", line 597, in interpret
    assert isinstance(expr1, ListNode)
AssertionError
```

ltratt:
[`rng3.eco` seems fine BTW.]

ltratt:
Should we merge this PR and fix the `range.eco` bug separately or ...?

ptersilie:
Nah I think I should fix that as well in here. This has to do with the AST that is apparently not properly initialised after loading a file. Since the other commits also fix stuff with file loading I think it would fit in here.

ltratt:
OK, sounds good to me. I won't merge yet :)

ptersilie:
I have reason to believe that that error only occurs because you have an outdated Eco file (i.e. the grammar has changed) which first results in an error and then on a reparse tries to reuse parse trees that are not valid anymore because the grammar has changed.

If you fix the file this shouldn't happen anymore, even if you save the file with parsing errors, as long as the parsing errors happened with the same grammar.

ptersilie:
I squashed. I think we can merge this. Let me know if you spot any more errors related to this.

ltratt:
@lukasd I think you're right. I've recreated range.eco from scratch, and the new version loads back in correctly.

ltratt:
OK, do we want to squash this before I merge?

ptersilie:
I think it's fine as it is.

PR #189, opened 2017-03-29
Speedup loading cached grammars

ptersilie:
This PR introduces an optimisation that I discovered while working on the auto language boxes. I realised that inserting new language boxes was quite slow (maybe half a second). Seeking to optimise this, I noticed that each time a new language box is created, the syntaxtable for that language is loaded from the disk. This of course is not needed if we already have that language loaded and only results in unnecessary slowdown on every additional language box we add to the program.

This PR optimises this by caching the syntaxtable the first time it is loaded and then shares that syntaxtable between all incremental parsers of the same language. This not only makes any additional language box to appear instantly, it also reduces the time of our tests suite from 90 to 37 seconds (on my machine).

The improvement can easily be tested by creating a Python + SQL program, and then adding multiple SQL language boxes. To optimise this further we could preload all valid language boxes when we create a new file, which would also get rid of the small lag when creating the first language box. Thoughts?

### DIFF
--- a/lib/eco/grammars/grammars.py
+++ b/lib/eco/grammars/grammars.py
@@ -52,26 +52,14 @@ class EcoFile(object):
     def load(self):
         from grammar_parser.bootstrap import BootstrapParser
         from jsonmanager import JsonManager
+        from incparser.incparser import IncParser
 
         if _cache.has_key(self.name + "::parser"):
 
-            root, language, whitespaces = _cache[self.name + "::json"]
-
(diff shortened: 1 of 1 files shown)


ltratt:
I think I'm fine with the lag for the first box, so let's keep this as-is. Nice work! Shall I merge now?

ptersilie:
Okay, in my auto language box branch I have to preload the grammars anyways, so as soon as we push that the first language box lag will be gone.

Yes, I think this is ready to merge.

PR #191, opened 2017-07-12
Implemented FuzzyReducer

ptersilie:
Helps to reduce the logs from failed fuzzy tests to a smaller size which makes them easier to debug and to create tests from them.

### DIFF
--- a/lib/eco/minimiser.py
+++ b/lib/eco/minimiser.py
@@ -1,0 +1,156 @@
+from grammars.grammars import lang_dict
+from treemanager import TreeManager
+from utils import KEY_UP as UP, KEY_DOWN as DOWN, KEY_LEFT as LEFT, KEY_RIGHT as RIGHT
+from optparse import OptionParser
+import logging
+import sys, os
+
+class FuzzyMinimiser(object):
+
+    def __init__(self, lang, program, test):
+        grm = lang_dict[language]
+        parser, lexer = grm.load()
+        parser.init_ast()
+        self.parser = parser
+        self.lexer = lexer
+        self.grm = grm
+        self.treemanager = TreeManager()
+        self.treemanager.add_parser(parser, lexer, grm.name)
+        self.treemanager.version = 1
+        self.treemanager.last_saved_version = 1
+
+        self.program = program
+        self.test = test
+        self.pos = 0
+        self.lastexception = None
+
+    def reset(self):
+        self.parser.reset()
+        self.treemanager = TreeManager()
+        self.treemanager.add_parser(self.parser, self.lexer, self.grm)
+
+    def auto(self):
+        logging.info("Running auto-mode")
+        # try to reduce snapshots first
+        while True:
+            self.original = self.test[:] # remember original tests
+            self.reduce("snapshot")
+            if self.test == self.original:
+                logging.info("No more snapshot reductions possible.")
+                break
+        # try single reduce next
+        while True:
+            self.original = self.test[:] # remember original tests
+            self.reduce("single")
+            if self.test == self.original:
+                logging.info("No more single reductions possible.")
+                break
+
+    def reduce(self, mode):
+        self.pos = 0
+        logging.info("Pos: {} Len: {} Start {}-reduce".format(self.pos, len(self.test), mode))
+        try:
+            # Run without changes to store exception
+            self.run()
+        except Exception as e:
+            self.lastexception = e
+        while self.pos < len(self.test):
+            self.delete(mode)
+            try:
+                self.run() # check if first run causes error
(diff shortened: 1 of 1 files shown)


lib/eco/reducer.py
```diff
@@ -0,0 +1,153 @@
...
+                    logging.info("Pos: {} Len: {} Still failing: '{}'".format(self.pos, len(self.test), e))
+                    self.save() # save progress
+        self.save()
+
+    def same_exception(self, e):
+        if not self.lastexception:
+            self.lastexception = e
+            return True
+        if type(self.lastexception) is type(e) and \
+                self.lastexception.args == e.args:
```
ltratt: Seems like a bit too much indentation here?

lib/eco/reducer.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def delete(self, mode):
+        self.old = self.test[:] # copy
+        if mode == "single":
+            del self.test[self.pos:self.pos+4] # remove one keystroke
+            if self.pos < len(self.test) and \
+                    self.test[self.pos] == "self.treemanager.undo_snapshot()":
+                        self.test.pop(self.pos)
+        elif mode == "snapshot":
+            del self.test[self.pos:self.pos+81] # remove one snapshot
```
ltratt: Is 81 an intentional constant? If so, should it be hoisted to a global (immutable) variable at the top of the file?

lib/eco/reducer.py
```diff
@@ -0,0 +1,153 @@
...
+    def run(self):
+        self.reset()
+        self.treemanager.import_file(self.program)
+        for l in self.test:
+            exec(l)
+
+    def move(self, direction, times):
+        for i in range(times): self.treemanager.cursor_movement(direction)
+
+optp = OptionParser(usage="usage: python2 %prog PROGRAM TEST LANG [options]\n\ne.g. python2 %prog program.txt tests.txt \"Python 2.7.5\"")
```
ltratt: Probably best to wrap this in a `if __name__ == "__main__"`, just in case you want to import this as a library one day.

ltratt (approved):

lib/eco/reducer.py
ptersilie: I just used the auto-indent feature from vim.

ltratt: Vim's autoindenter is sometimes broken in my experience :( Best to fix this one by hand.

lib/eco/reducer.py
ptersilie: Yes it is intentional since each snapshot contains 20 keystrokes where 1 keystroke consists of 4 lines plus an additional `save_snapshot` line. I don't mind moving it into a global variable tho.

lib/eco/reducer.py
ptersilie: Fair point.

lib/eco/reducer.py
ltratt: Alternatively, you could explain in a comment why 81 is 81 (it's an unusual number in a program ;) ).

ptersilie:
Can I squash this?

ltratt:
Yep!

ptersilie:
Actually, I think you can just use the squash/merge button, as I have nothing else to change.

ltratt:
It's best if you squash it, so you know what the commit message will be.

ptersilie:
Squashed!

ptersilie:
Ah wait. I was just wondering, if `FuzzyMinimiser` is not a better name! What do you think?

ltratt:
Yep, that's probably better. Feel free to fix.

ptersilie:
Okay, renamed it.

ltratt:
Would it be easiest to just call the file `minimiser.py`? In a sense, the fact that it uses fuzziness internally is less relevant than the fact that it's a minimiser. The name of the class inside seems sensible to me though.

ptersilie:
I just thought I might have another minimiser in the future.

ltratt:
I guess I thought it would be a flag to this program (e.g. `minimiser.py -backend fuzzy` or whatever). I don't feel strongly about it, so if you'd prefer to leave things as they are, no problem.

ptersilie:
What I meant is that I thought I might have a parsing minimiser or something similar later on. But to be honest I'm not fussed. I did name it `minimiser.py` at first, so I am not opposed to the idea. I will just change it. I think it's unlikely I'll have other minimisers anyways, and if so, possibly in a subfolder.

ltratt:
Looks good. Time to squash?

ptersilie:
Ready to merge!

PR #193, opened 2017-07-19
Fixed bug where deleting a selection is really slow

ptersilie:
When incrementally relexing a node, the algorithm stops when a token is relexed to the same value; only, however, if we passed the initial node where relexing started. There was a bug that when the initial startnode was merged with it's left neighbor, the algorithm wouldn't notice that we passed the startnode and thus relex the entire file. This fix alters the check so that a merged startnode also triggers the `past_startnode` flag.

### DIFF
--- a/lib/eco/inclexer/inclexer.py
+++ b/lib/eco/inclexer/inclexer.py
@@ -515,8 +515,11 @@ class IncrementalLexerCF(object):
             try:
                 token = next_token()
                 lookaheads.append(token[2])
-                if token[3][0] is startnode:
-                    past_startnode = True
+                if not past_startnode:
+                    for temp in token[3]:
+                        if temp is startnode:
+                            past_startnode = True
+                            break
                 toks.append(token[:3])
                 tokenslength += len(token[0])
                 for r in token[3]:


lib/eco/inclexer/inclexer.py
```diff
@@ -515,8 +515,10 @@ def relex(self, node):
             try:
                 token = next_token()
                 lookaheads.append(token[2])
-                if token[3][0] is startnode:
-                    past_startnode = True
+                for temp in token[3]:
```
ltratt: Is it just me or is this loop only worth taking if `past_startnode` is `False`? [I don't know if the loop is over a long data-structure or not; if it is potentially long, adding `if not past_startnode` might be a useful optimisation.]

ptersilie: It's probably one to two elements most of the time. Worst case scenarios are huge strings.

ltratt: So better left as-is or ... ? [I don't mind, but just want to be sure you've thought about it!]

ptersilie: But I guess it can't hurt to put that extra statement in.

ptersilie: Nah I changed it now. I think it makes sense to have it in their in case there ever is a worst case scenario. The check is so cheap there's no reason not to put it in.

ltratt:
OK. Squash and then I can merge?

ptersilie:
Squashed.

PR #194, opened 2017-09-01
Fix PyQt4 dependency on travis

ptersilie:
Since travis switched to Ubuntu Trusty we can't install dependencies via apt and use them with `system_site_packages` anymore. Since our tests depend on PyQt4 we have to manually build and install it ourselves.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,17 +4,19 @@ python: "2.7"
 virtualenv:
   system_site_packages: true
 
-addons:
-  apt:
-    packages:
-      python-qt4
-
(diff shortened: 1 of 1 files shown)


ltratt:
All a bit "eek", but well done for finding a solution to such an annoying problem.

ptersilie:
I think PyQt5 can be installed via pip so if we upgrade to python3 and PyQt5 one day we can revert this back.

ltratt:
As things stand, I don't think we'll need to worry about that ;)

PR #197, opened 2017-10-10
Fix pygame bug when drawing parse trees

ptersilie:
It is recommended to draw rectangles using surface.fill(). However, there seems to be a bug where those rectangles can't be drawn outside of the viewport and instead are drawn at position 0. Switching to  draw.rect() fixes this.

You can check this by inspecting a parse tree and then moving the view around. The white background rectangles of the tree nodes won't move outside the view prior to this fix. I upgraded my pygame version thinking this bug might have been fixed but it still seems to be there.


ptersilie:
I should have put this into the improvements PR but forgot about it.

PR #198, opened 2017-10-18
Retainbug

ptersilie:
This PR fixes two bugs found by the fuzzy tester. Might need rebasing, as I'm unsure if I should keep the commits separate or not.

### DIFF
--- a/lib/eco/incparser/astree.py
+++ b/lib/eco/incparser/astree.py
@@ -224,8 +224,12 @@ class Node(object):
             if c is self:
                 if i > 0:
                     c.left = self.parent.children[i-1]
+                else:
+                    c.left = None
                 if i < len(self.parent.children) - 1:
                     c.right = self.parent.children[i+1]
+                else:
+                    c.right = None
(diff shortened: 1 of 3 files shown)


lib/eco/incparser/astree.py
```diff
@@ -222,10 +222,15 @@ def update_siblings(self):
         for i in range(len(self.parent.children)):
             c = self.parent.children[i]
             if c is self:
+
```
ltratt: Random blank line?

ltratt:
One minor comment to address. The two commits should be separate IMHO (after the comment commit has been squashed away).

lib/eco/incparser/astree.py
ptersilie: fixed

ltratt:
OK, if you squash that last commit into whichever of the other two it should be part of, then I think I can merge.

ptersilie:
Squashed!

ltratt (approved):

PR #201, opened 2017-12-06
Fixed retainability check

ptersilie:
A node can be retained if its content has the same textlength and the node
has the same position as before parsing. Previously we only checked
that the textlength matches. This commit adds the position check to
the retainability query.

### DIFF
--- a/lib/eco/incparser/astree.py
+++ b/lib/eco/incparser/astree.py
@@ -40,55 +40,6 @@ class AST(object):
     def get_bos(self):
         return self.parent.children[0]
 
-    def get_nodes_at_position(self, pos, a=None, b=None):
-        """
-        Searches all nodes that match the current cursor position in the TextField.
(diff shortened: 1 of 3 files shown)


ltratt:
This looks good. Do you want to squash before I merge?

ptersilie:
I wasn't planning to but don't mind it either. You could also use Github's squash and merge feature.

ltratt:
[We don't use GH's squash feature: it's the PR author's job to squash (or not).]

I think you might want to quickly take a look and see if all three commits are meaningful in their own right. If you think they are, I'll merge as-is.

ptersilie:
Ye I think they are meaningful on their own which is why I wasn't planning on merging them. But I'm also not opposed to it. Personally I would merge as is.

ltratt:
Merge as-is it is!

PR #203, opened 2018-01-09
Customlexer

ptersilie:
This PR adds our own custom lexer to Eco, which allows us to handle language boxes properly without having to revert to hacks. This also solves the obvious problems that arise with using `\x80` as a placeholder for language boxes in strings.


ltratt:
If you can squash some of the rubbish commits away before review, that would be helpful :)

Do you have a rough benchmark of before/after performance?

ptersilie:
Squashed away the rubbish commits. Also compared the running time of the tests and there's a roughly 9% slowdown with this PR (89 seconds vs 81 seconds).

ltratt:
OK, this isn't going to be a quick review, because of the size, so it'll take me until well in to tomorrow to start it. Sorry!

ptersilie:
No worries, I'll have to fix another bug I found first anyways. So it's not urgent.

ltratt:
Ah, OK. Let me know when that's done.

lib/eco/treelexer/lexer.py
```diff
@@ -1,3 +1,29 @@
+class RE_OR(object):
+    def __init__(self, lhs, rhs):
+        self.lhs = lhs
+        self.rhs = rhs
+
+    def __repr__(self):
+        return "OR(%s, %s)" % (self.lhs, self.rhs)
+
+    def __eq__(self, other):
+        return type(other) is type(self) and \
```
ltratt, line 10: Do you need the `type` checks? At least you could simplify them to `isinstance(other, RE_OR)`?

lib/eco/treelexer/lexer.py
```diff
@@ -42,3 +68,68 @@ def match(self, pattern, text):
                 return self.match_star(pattern[0] + "*" + pattern[2:], text[1:])
             return False
         return self.match_one(pattern, text) and self.match(pattern[1:], text[1:])
+
+class RegexParser(object):
+
```
ltratt: Blank line.

lib/eco/treelexer/lexer.py
```diff
@@ -42,3 +68,68 @@ def match(self, pattern, text):
...
+    def __init__(self):
+        from grammars.grammars import regex
```
ltratt: Why not import at the top level?

lib/eco/treelexer/lexer.py
```diff
@@ -42,3 +68,68 @@ def match(self, pattern, text):
...
+        elif node.name == "STAR":
+            return self.parse_star(node)
+        elif node.name == "PLUS":
+            return self.parse_plus(node)
+        elif node.name == "CHAR":
+            return self.parse_char(node)
+
+    def parse_list(self, node):
+        if len(node.children) > 1:
+            l = []
```
ltratt: Could be `return [self.parse(c) for c in node.children]` I think.

lib/eco/treelexer/test_lexer.py
```diff
@@ -1,4 +1,32 @@
-from lexer import PatternMatcher
+from lexer import PatternMatcher, RegexParser, RE_CHAR, RE_OR, RE_STAR, RE_PLUS
+
+class Test_RegexParser(object):
+
```
ltratt: Blank line.

lib/eco/treelexer/test_lexer.py
```diff
@@ -1,4 +1,32 @@
...
+        assert self.regparse.ast("ab") == [RE_CHAR("a"), RE_CHAR("b")]
+        assert self.regparse.ast("a|b") == RE_OR(RE_CHAR("a"), RE_CHAR("b"))
+        assert self.regparse.ast("a*") == RE_STAR(RE_CHAR("a"))
+        assert self.regparse.ast("a+") == RE_PLUS(RE_CHAR("a"))
+        assert self.regparse.ast("(a)") == RE_CHAR("a")
+        assert self.regparse.ast("(ab)") == [RE_CHAR("a"), RE_CHAR("b")]
+        assert self.regparse.ast("(ab)*") == RE_STAR([RE_CHAR("a"), RE_CHAR("b")])
+
+        assert self.regparse.ast("(a|b)*|c+") == RE_OR(\
+                                             RE_STAR(\
```
ltratt: The indentation here looks weird?

lib/eco/treelexer/lexer.py
```diff
@@ -49,25 +56,50 @@ def match_question(self, pattern, text):
...
+        while True:
+            if not self.match(pattern.c, text):
+                break
+        return True
+
+    def match_plus(self, pattern, text):
+        if not self.match(pattern.c, text):
+            # we have to at least match one
+            return False
+        while True:
```
ltratt: `while self.match(...): pass` is equivalent I think?

lib/eco/treelexer/lexer.py
```diff
@@ -49,25 +56,50 @@ def match_question(self, pattern, text):
...
+        return True
+
+    def match_list(self, pattern, text):
+        for p in pattern:
+            if not self.match(p, text):
+                return False
+        return True
+
+    def match_or(self, pattern, text):
+        tmp = self.pos
```
ltratt: Ah, I wondered when doing `self.pos` was going to bite you ;)

I'm not sure it's worth changing things now, but I *much* prefer passing and returning `pos` as a parameter, so that you can "undo" it without having to do global mutation. [This is how things like lrtable work FWIW.]

lib/eco/treelexer/lexer.py
```diff
@@ -49,25 +56,50 @@ def match_question(self, pattern, text):
...
+        if self.match(pattern.lhs, text):
+            return True
+        self.pos = tmp # backtrack
+        return self.match(pattern.rhs, text)
 
     def match(self, pattern, text):
+        print("match", pattern, text)
```
ltratt: Stray debugging statement?

lib/eco/treelexer/lexer.py
```diff
@@ -83,6 +84,17 @@ def match_or(self, pattern, text):
         self.pos = tmp # backtrack
         return self.match(pattern.rhs, text)
 
+    def match_range(self, pattern, text):
+        if self.pos >= len(text):
+            return False
+
+        s = ord(pattern.c[0])
+        e = ord(pattern.c[2])
```
ltratt: Do you check that `s` is less than `e` somewhere? If not, add an `assert` here (otherwise you're going to get weirdness if you do `[z-a]`.

lib/eco/treelexer/test_lexer.py
```diff
@@ -112,5 +112,24 @@ def test_escaped(self):
...
+        assert PatternMatcher().match(self.cmp("[a-zA-Z_][a-zA-Z_0-9]*"), "123abc123_") is False
+
+        p = PatternMatcher()
+        assert p.match(self.cmp("#[^\\r]*"), "# abc") is True
+        assert p.pos == 5
+
+        p = PatternMatcher()
+        assert p.match(self.cmp("#[^\r]*"), "# abc \r") is True
+        assert p.pos == 6
+
```
ltratt: Double blank line.

lib/eco/treelexer/lexer.py
```diff
@@ -215,3 +218,24 @@ def parse_plus(self, node):
     def parse_question(self, node):
         content = self.parse(node.get("value"))
         return RE_QUESTION(content)
+
+class Lexer(object):
+
```
ltratt: Blank line.

lib/eco/treelexer/test_lexer.py
```diff
@@ -133,3 +134,10 @@ def test_realworld_examples(self):
         assert PatternMatcher().match(self.cmp("([0-9]+\.?[0-9]*|\.[0-9]+)([eE](\+|-)?[0-9]+)?"), "1e23") is True
         assert PatternMatcher().match(self.cmp("\'[^\'\r]*\'"), "'this is a string 123!'") is True
         assert PatternMatcher().match(self.cmp("\'[^\'\r]*\'"), "'this is a with a newline \r string 123!'") is False
+
+class Test_Lexer(object):
+
```
ltratt: Blank line.

lib/eco/treelexer/lexer.py
```diff
@@ -34,9 +34,11 @@ class PatternMatcher(object):
 
     def __init__(self):
         self.pos = 0
+        self.exactmatch = True
```
ltratt: What does this attribute mean?

lib/eco/treelexer/lexer.py
```diff
@@ -229,13 +233,24 @@ def __init__(self, rules):
             self.patterns.append((pattern, name))
 
     def lex(self, text):
+        """Lexes a given string by trying all patterns. When a match is found a
+        token is created, and the lexing continues with the remainder of the
+        string. Lookahead is calculated by observing how far the patternmatcher
+        got before returning 'no match'. If the pattern is open ended, e.g. +,
+        *, etc, the match is not exact and we need to add 1 to the lookahead,
```
ltratt: Not sure what "need to add 1" means... Is it a fixed thing or infinite or ...?

lib/eco/treelexer/lexer.py
```diff
@@ -132,6 +120,12 @@ def match(self, pattern, text):
             return self.match_question(pattern, text)
         raise NotImplementedError(pattern)
 
+    def match(self, pattern, text):
```
ltratt: I'm not a fan of having `match` and `_match` -- that's just too close a typo to make. Can we rename one method to something a bit more obviously distinct?

lib/eco/treelexer/lexer.py
```diff
@@ -257,3 +309,27 @@ def lex(self, text):
...
+        result = []
+        while True:
+            lookahead = 0
+            oldpos = pos
+            oldnode = node
+            for p, n in self.patterns:
+                pm.reset()
+                token = pm.match(p, node)
+                if token:
+                    lookahead = lookahead + (1 if not pm.exactmatch else 0)
```
ltratt: Let's please not use Python's awful ternary operator (which isn't left-to-right evaluated) -- an If statement would only be 1 line longer ;)

lib/eco/treelexer/test_lexer.py
```diff
@@ -43,7 +43,7 @@ def test_match_one(self):
         assert PatternMatcher().match(RE_CHAR("a"), "a") == "a"
         assert PatternMatcher().match(RE_CHAR("."), "c") == "c"
         assert PatternMatcher().match(RE_CHAR("x"), "c") is None
-        assert PatternMatcher().match(None, "c") == "c"
+        #assert PatternMatcher().match(None, "c") == "c"
```
ltratt: Commented code.

lib/eco/treelexer/test_lexer.py
```diff
@@ -208,3 +210,82 @@ def test_nodes(self):
 
         l = Lexer([("name", "[a-z]+")])
         assert l.treelex(a) == [("ab", "name", 0)]
+
+class Test_IncrementalLexing(object):
+
```
ltratt: Blank line.

lib/eco/treelexer/lexer.py
```diff
@@ -181,18 +181,22 @@ def append(self):
             self.result.append(self.char())
 
     def get_token(self):
```
ltratt: Can we document this method? Its name doesn't seem to quite match its functionality?

lib/eco/inclexer/inclexer.py
```diff
@@ -24,7 +24,7 @@
 from incparser.astree import BOS, EOS, TextNode, ImageNode, MultiTextNode
 from PyQt4.QtGui import QImage
 import re, os
-from cflexer.lexer import LexingError
+#from cflexer.lexer import LexingError
```
ltratt, line 27: Commented code.

lib/eco/treelexer/test_lexer.py
```diff
@@ -339,3 +363,15 @@ def test_token_iter_newline_lbox(self):
         assert it.next() == (["'a", "\r", "b", "c'"], "string", 1, [TextNode(Terminal("'a")), TextNode(Terminal("\r")), TextNode(Terminal("b")), TextNode(MagicTerminal("<SQL>")), TextNode(Terminal("c'"))])
         with pytest.raises(StopIteration):
             it.next()
+
+class Test_Keyword(object):
+
```
ltratt: Blank line.

lib/eco/treelexer/lexer.py
```diff
@@ -159,9 +159,17 @@ def match(self, pattern, text, pos=0):
             return self.get_token()
         return None
 
-from incparser.astree import TextNode, BOS, EOS
+from incparser.astree import TextNode, BOS, EOS, MultiTextNode
 from grammar_parser.gparser import MagicTerminal, Terminal, IndentationTerminal
 
+class LPH(object):
```
ltratt: Shouldn't this be `LBPH`?

ltratt:
I have a number of small comments, though nothing major. You probably need to expand some of those supposedly "out of date" commits to see if they're still relevant or not.

Two overall comments.

1) You don't want to use `self.pos` for hand-rolled parsers -- as you've found out, it doesn't play nicely with backtracking :) I just pass a variable `pos` as a parameter to functions and return it on success, which naturally models backtracking.

2) Can we expand the commit messages? They're often too terse for me to really work out what's going on. You don't need to write essays, but it's important to give the reviewer (and later reader of history) a sense of not just what you've done but why. Sometimes 1 sentence extra is all that's necessary (occasionally you'll need much more; most of the time a short paragraph is the happy medium).

lib/eco/treelexer/lexer.py
ptersilie: I started to use type instead of isinstance, as it is more precise and possibly faster as it doesn't need to check for superclasses.

ltratt: Good point about performance... But do you even need the `type` checks at all?

ptersilie: I think I initally wrote it this way because the `__eq__` method used to be shared among multiple classes before I realised I had to customise the method anyways. So I could just say `type(other) is RE_OR`.

lib/eco/treelexer/lexer.py
ptersilie: True.

lib/eco/treelexer/lexer.py
ltratt: Up to you.

ptersilie: I do need the type checks as I don't want `RE_OR` and `RE_PLUS` to be equivalent. I don't think it's worth changing it to `type(other) is RE_OR`. Also as it is now it matches the style `__eq__` in `RE_DEFAULT` where the `type(other) is type(self)` makes sense.

ltratt: Fair enough.

lib/eco/treelexer/lexer.py
ptersilie: This attribute is needed for lookahead calculation. It tells us if a token was matched by something like `[a-z]+`. In that case the match is not exact and token needs a lookahead of 1. On the other hand matching `abc` is exact as there is no longer match possible and thus the lookahead would be 0.

lib/eco/treelexer/lexer.py
ptersilie: Indeed. I didn't remember that `or` introduces backtracking. You'll see in a later commit that I need to store even more state, so at least it prepared me for that. ;-)

lib/eco/treelexer/lexer.py
ltratt: Ah. Should it have a default value?

lib/eco/treelexer/lexer.py
ptersilie: If the token was matched by `+` or `*` then it always has a lookahead of at least 1. If it already has a lookahead > 1 then it's lookahead is 1 more as it needs to look at 1 additional character to know that there is no longer match.

lib/eco/treelexer/lexer.py
ptersilie: I guess I tried to keep it 3 chars, but this is also fine.

lib/eco/treelexer/lexer.py
ptersilie: Sure I couldn't think of anything tho. `_match` used to be `match` before I realised I needed another method to be called `match`. I welcome any suggestions.

lib/eco/treelexer/lexer.py
ltratt: OK. Maybe a quick rephrasing might help here?

lib/eco/treelexer/lexer.py
ptersilie: Yes by default matches are exact until a match fails (a character was found that doesn't match which means we return what we've found so far and set the lookahead to 1).

lib/eco/treelexer/lexer.py
ptersilie: No here, but one of the later commits checks for this.

lib/eco/treelexer/lexer.py
ptersilie: Probably also removed in the later commits.

lib/eco/treelexer/lexer.py
ltratt: Nearly anything'll be better ;) Even `_inner_match` or something like that.

lib/eco/treelexer/lexer.py
ltratt: OK.

lib/eco/treelexer/lexer.py
ltratt: Probably? ;)

ptersilie:
I think I have addressed all your comments. I am not sure how to squash the last commit though as it changes various commits from this PR.

ltratt:
These look good to me.

What I would sometimes do with things like that last commit is to break it up into the necessary sub-commits, then squash each as necessary. That said, I almost never do a single "address comments" single commit, to avoid having to split it up later ;)

ptersilie:
I will try. It seems like a tedious job for little gain though.

ltratt:
I'm not going to make you do it, though it is excellent git practise!

ptersilie:
I managed to split up some commits in the history again (a first for me) and rebase as much as I could. The last commit "rename _match" is a bit tricky as rebasing this with "Patternmatcher returns matched..." will cause conflicts in pretty much every commit following this, so I'm leaving it in.

ltratt:
Looks good to me (and the rename is fine). Shall I merge as-is (once the tests have run)?

ptersilie:
Sure. What this "Update branch" button? I've never seen that before.

ltratt:
Ah, you haven't rebased against `origin/master`. If you do that and force push, that button should disappear.

ptersilie:
Do you wanna try using that button instead to see what it does?

ltratt:
Nope, I don't want a merge commit ;) Basically your final rebase should always be against `origin/master`, so this is something you need to do.

ptersilie:
Fair enough

ptersilie:
And done!

ltratt (approved):

ltratt:
Since https://github.com/softdevteam/eco/pull/206 is merged, you'll have to rebase against `origin/master` (after doing `git fetch`) again. Sorry :/

ltratt:
Ping.

ptersilie:
This is ready for further review again.

ltratt:
OK I think I can merge this one as-is now?

ptersilie:
If you have no further comments, then go ahead.

PR #206, opened 2018-01-11
Cleaned up test suite

ptersilie:
In light of the problems with travis I deactivated the slow running tests
and we will run our own cronjob instead. I also took the opportunity to
clean up the test suite by removing some deprecated tests and adding more
tests to be run by default.


lib/eco/incparser/test/test_lrparser.py
```diff
@@ -30,7 +30,7 @@
 """
 
 import pytest
-pytest.skip("deprecated")
+pytestmark = pytest.mark.skip
```
ltratt: What's the point of this?

ptersilie: This test is deprecated. I could also just delete it I guess. Wasn't sure if this might be useful again some day.

ltratt: Up to you. I'm generally not a fan of keeping dead code around unless there's a very good reason for doing so. Dead code is always in the git history if it needs to be fished out...

ptersilie: Fair enough. I'll remove it.

ptersilie:
Ready to squash?

ltratt:
OK.

ptersilie:
Done.

ltratt (approved):

PR #208, opened 2018-02-13
Properly set exists flags and calculate node position

ptersilie:
This commit fixes a few inconsistencies with the node.exists-flag which can cause problems with the retain algorithms. This commit also fixes the position calculation used for same_text_pos which was inaccurate. We can't use the position from the history as changing a token at the beginning of the file we'd have to update the positions of all other tokens in the entire tree. Luckily, we only need the position within the `same_text_pos` function. When we reach that function we will have visited all changed nodes up to that point so we can calculate their position on the fly during `pass1`.

### DIFF
--- a/lib/eco/incparser/astree.py
+++ b/lib/eco/incparser/astree.py
@@ -584,10 +584,6 @@ class TextNode(Node):
             return self.get_attr("changed", version) or self.get_attr("nested_changes", version)
         return self.changed or self.nested_changes
 
-    def does_exist(self):
-        # XXX isn't it also either new or reused?
-        return self.exists
-
     def has_errors(self):
         return self.nested_errors or self.local_error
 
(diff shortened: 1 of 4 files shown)


ltratt:
Dumb question: I think we can delete the `does_exist` function entirely as this was the only call site?

ltratt:
I kicked the build to no avail. I think this is why:

```
$ wget https://sourceforge.net/projects/downloads/sip.tar.gz
$ cat sip.tar.gz
...
We're sorry -- the Sourceforge site is currently in Disaster Recovery mode, and currently requires
the use of javascript to function.  Please check back later.
...
```

Probably worth kicking it again tomorrow in case this problem has resolved itself.

ltratt:
Aha, it looks like sourceforge is back up now.

ptersilie:
Yes, `does_exist` can go of course!

ltratt:
If you squash, I can merge.

ptersilie:
Squashed!

ltratt (approved):

PR #210, opened 2018-03-02
Fix Java lexing rules for integers and floats

### LINKED ISSUE #209: Java lexing bug
This (correct) Java file doesn't lex / parse properly in Eco:

```java
class C {
    float f = 1.0F;
}
```

because of the `1.0F` part.

ptersilie:
Fixes #209


ltratt:
[One build died, but for silly looking reasons. I've kicked it.]

ltratt:
Hmm, both builds are saying `The command "tar xzf downloads/sip.tar.gz -C downloads --keep-newer-files" failed and exited with 2 during .`. Any thoughts?

ptersilie:
Oh I think sourceforge is down again. :/ I'm still annoyed that we are dependent on that.

ltratt:
Looks like it worked. Squash then I can merge?

ptersilie:
Sure. After I squashed let me first check if the caching works.

ltratt:
Did it work?

ptersilie:
Not really. I added another commit to see if that keeps more files so that we don't have to build sip and pyqt each time the tests run.

ptersilie:
Haha, the mirrorservice we were using for sourceforge is down as well now. :/

ltratt:
The mirrorservice link works for me. I've tried kicking the build...

ltratt:
OK, so the download did work. But now we have `tar` dying:

```
tar xzf downloads/sip.tar.gz -C downloads --keep-old-files
tar: sip-4.19.3/ChangeLog: Cannot open: File exists
tar: sip-4.19.3/configure.py: Cannot open: File exists
...
```

and so on.

I suggest we probably want to do something like:

```
test -f sip-4.19.3/ChangeLog || tar xzf downloads/sip.tar.gz -C downloads --keep-old-files
```

or just remove the `--keep-old-files`.

ptersilie:
Oh I thought the `keep-old-files` would take care of the `File exists` by just "keeping the old files". What I want is to not overwrite the files again so `make` will not rebuild everything again. Is that what your `test` command does?

ltratt:
What my suggestion does is not even run `tar` if (at least one of) the files exists.

ptersilie:
Hmm, even though I'm caching the build folders now and not overwriting the files again tar, PyQt still rebuild every time travis runs, which takes 4x longer than the tests.

ltratt:
You probably want to guard these lines:

```
     cd downloads/sip-4.19.3 && python configure.py && sudo make && sudo make install && cd ../../
     cd downloads/PyQt4_gpl_x11-4.12.1 && python configure.py --confirm-license && sudo make && sudo make install && cd ../../
```

since, I think, at the moment they're run even if we haven't downloaded anything. The configure script probably touches some files which then cause make to do stuff.

ltratt:
Time to resurrect this one :)

ptersilie:
I added some more guards to avoid unneccesary building which seems to work now. If you are happy with this I can squash and we can merge.

ptersilie:
This reduced the test runtime from 18 to 3 minutes.

ltratt:
> This reduced the test runtime from 18 to 3 minutes.

Eh?! Any idea why?

ptersilie:
Because it doesn't need to rebuild PyQt each time which was really slow. The whole reason I did all those experiments.

ltratt:
Aha, I see. So is this ready for squashing?

ptersilie:
Yes, I can squash if you are happy.

ltratt:
Yep, please squash.

ptersilie:
Squashed!

ltratt (approved):

PR #212, opened 2018-06-26
Limit automatic language box detection

ptersilie:
This PR adds options to limit the automatic detection of language boxes to specific tokens. This is used
to avoid that compositions with HTML detect boxes everywhere, since any text is valid in HTML.

One remaining question is, if we should reverse the way these tokens are defined, e.g. instead of specifying what CAN be allowed, we need to specify what CANNOT be allowed. For HTML this would make defining the composition a lot easier as we could just disallow the `TEXT` token. However, for other languages it might be easier to define what is allowed.

### DIFF
--- a/lib/eco/grammars/grammars.py
+++ b/lib/eco/grammars/grammars.py
@@ -49,6 +49,8 @@ class EcoFile(object):
         self.alts = {}
         self.included_langs = set()
         self.extract = None
+        self.auto_include = None
+        self.auto_exclude = None
 
     def load(self, buildlexer=True):
         from grammar_parser.bootstrap import BootstrapParser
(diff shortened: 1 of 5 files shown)


lib/eco/grammars/grammars.py
```diff
@@ -103,6 +104,17 @@ def add_alternative(self, nonterminal, language):
     def change_start(self, name):
         self.extract = name
 
+    def set_auto_limit(self, lang, tokentype):
+        if self.auto_limit_start is None:
+            self.auto_limit_start = {}
+        self.auto_limit_start[lang] = tokentype
+
+    def auto_allows(self, lang, tokentype):
```
ltratt: This function returns `True` or `None` which seems weird? Could we simplify it to:

```python
def auto_allows(self, lang, tokentype):
  if self.auto_limit_start and self.auto_limit_start.has_key(lang):
    return tokentype in self.auto_limit_start[lang]
  return False
```

or similar?

ltratt:
I don't think it has to be either/or does it? We could have options to both be inclusive or exclusive when composing?

ptersilie:
I was thinking about that. We certainly can, although it would allow us to define contradicting rules, e.g. allow everything but `<html>` and don't allow `<html>`.

ltratt:
True, although we could probably detect that easily enough statically?

ptersilie:
Added exclusion rules and some tests

lib/eco/grammars/grammars.py
ltratt: With a little updating, I think the above suggestion can be made to work in the code below.

lib/eco/grammars/grammars.py
```diff
@@ -104,16 +105,31 @@ def add_alternative(self, nonterminal, language):
...
-            self.auto_limit_start = {}
-        self.auto_limit_start[lang] = tokentype
+    def set_auto_include(self, lang, tokentype):
+        if self.auto_include is None:
+            self.auto_include = {}
+        self.auto_include[lang] = tokentype
+
+    def set_auto_exclude(self, lang, tokentype):
+        if self.auto_include is not None:
+            print "Warning: Inclusion and exclusion rules may conflict! Exclusion rules are ignored."
```
ltratt: "Exclusion rules will be ignored"

ptersilie: Fixed!

ptersilie:
Fixed! Squash?

ltratt:
Please squash.

ltratt (approved):

PR #215, opened 2018-08-24
Fix regex parser

ptersilie:
Added support for non-greedy star and fixed greedy star.


lib/eco/treelexer/lexer.py
```diff
@@ -80,24 +88,54 @@ def match_one(self, pattern):
...
+            while True:
+                tmp = self.save_state()
+                if self.match_list(l[1:]):
+                    return True
+                self.load_state(tmp)
+                if self._inner_match(pattern.c):
+                    continue
+                return False
+        else: # greedy
+            tmp = self.save_state()
```
ltratt: I might be wrong, but I don't see how this backtracks more than one level? I was expecting a loop here. Can it deal with cases like `a*aa`?

ptersilie:
Ignore the first commit as that was a failed attempt. Actually, let me quickly rebase the first commit away, as that makes this easier to review.

ltratt:
Can you look at my comment before you do that?

ltratt:
If you force push, the comment might go away.

lib/eco/treelexer/lexer.py
ptersilie: It does, see new commit.

ptersilie: `match_list` is recursive and will call `match_star` again, which makes the backtracking work. I tried to do this with a normal loop but failed.

lib/eco/treelexer/lexer.py
```diff
@@ -80,24 +88,54 @@ def match_one(self, pattern):
...
+            if self._inner_match(pattern.c) and self.match_list(l):
+                return True
+            # remember how far we looked ahead before backtracking
+            etmp = (self.la, self.exactmatch)
+            self.load_state(tmp)
+            self.la, self.exactmatch = etmp
+            if self.match_list(l[1:]):
```
ltratt: Ok, so this bit is confusing. I expected the greedy backtracing to count down negatively. So let's say we greedily matched 5 characters, we'd then backtrack and try 4, then 3 (etc.). This looks like it's counting up from 0?

ptersilie: This only removes the star regex from the list and matches the rest. E.g. if the regex is `a*aa`, it first matches greedily, which then fails above so we backtrack, remove `a*` and match the remaining `aa`.

ltratt: That doesn't sound right. How can we still match an arbitrary number of `a*`s? It sounds like we match "all" or "none" (or maybe one?)?

ptersilie: If you want to give me a test I can check if it fails.

ptersilie: Note that line 108 recursively matches star again as it is contained within `l`. So it says something like `if match(a*) and match(a*aaa)`

ltratt: Can you be a bit clearer what the `l[1:]` bit is doing? I'm not sure I've understood either of the explanations so far correctly.

ptersilie: It removes the star regex from the list of regex's it still has to match. Let's say `match_star` starts with something like `a*aa`. Then `pattern.c` is `a*`, while `l` is the entire regex `a*aa`. We then try to recursively do `match_star(a*) and match(a*aa)`, which greedily matches as many `a`'s as it can until it fails because there are no more `a`'s to match. This means it has matched too many `a`'s using `a*` so there is nothing left to match the remaining bit `aa`. So it backtracks until it can match `l[1:]` which is just `aa`.

I agree this is a bit confusing due to the somewhat hidden recursion, but I couldn't find an easier way of doing this.

ltratt: OK, so I get the `pattern.c` thing now; but I don't see where the backtracking happens?

ptersilie: The backtracking happens when the remainder of the pattern `l` can't be matched in line 114. This then returns False, which makes `match(l)` 108 fail, which then backtracks pattern.c once (i.e. one `a`) and then tries to match `aa` again, if that still fails, it returns false, which backtracks another `a` and so on, until line 114 succeeds.

ltratt: Maybe I'm dumb, but backtracing should involve a `- 1` somewhere, I would assume? And I just don't see that. I feel like I'm missing out on something obvious.

ptersilie: Oh I see what you mean now! This happens within `self.load_state(tmp)`, which resets the `self.pos` (the position how far we've matched) to before a single `a` was matched. Due to the recursion this then happens multiple times until we are back at a `pos` where `match(l[1:])` can succeed.

ltratt: OK, so I feel a bit dumb: `load_state` doesn't fully reset the lexing state back to what it was at `save_state`?

ptersilie: It does (it resets `self.pos`), however that also resets the information about how far we had to look ahead before backtracking and if the match is exact, so I keep that information in Line 113.

ltratt: I think I might get it. If I have a regex `a*a` and the input "aaa", the "a*" will recurse three times, finish, try matching an a, fail, unwind one level of recrusion, and then try matching an a. In other words, the stack has a worst case depth of O(n) where *n* is the length of the input?

ptersilie: Yes, that sounds correct.

ltratt: OK: we can't do this, I'm afraid, because users will write perfectly normal content (e.g. a long string) that will blow up the stack. [I originally thought your recursion depth was O(m) where *m* is the number of elements in the regex. That's a standard-ish way of implementing regexs.]

ptersilie: I agree, I tried to make this non-recursive but gave up after a while. I'll have another jab at this, but if it takes too long I'll fix it it another PR.

ptersilie: Okay, have a look at the latest commit, which is a non-recursive solution for this.

ptersilie:
Okay, have a look at the latest commit, which is a non-recursive solution for greedy star.

ltratt:
We have failing tests...

ptersilie:
Just saw. :( Looking now

ptersilie:
Pushed some more changes fixing matchstar and also lookahead calculation.

lib/eco/treelexer/lexer.py
```diff
@@ -80,24 +89,59 @@ def match_one(self, pattern):
...
+                self.la = max_la
+            return False
+
+    def match_plus(self, pattern, l=None):
         if not self._inner_match(pattern.c):
             # we have to at least match one
             return False
-        return self.match_star(pattern)
+        if l:
+            # convert RE_PLUS to RE_STAR
```
ltratt: Why?

ltratt:
I think I believe this more now. Squash?

lib/eco/treelexer/lexer.py
ptersilie: This is no longer needed and I removed it in the latest commit.

ptersilie:
Just had to do one last commit, if you want to check it out. Then I can squash!

ltratt:
So that code is dead?

ptersilie:
Yes

ltratt:
OK, please squash (and rebase).

ptersilie:
Squashed!

ltratt:
You haven't rebased against `origin/master`.

ptersilie:
Whoops. Fixed! Probably have to flush the cache on Travis again as well.

ltratt:
OK, I'll wait until you've kicked the tests into life again.

ltratt (approved):

PR #218, opened 2018-09-04
Fix incremental AST reusing of nodes

ptersilie:
The AST consists of AstNode's and ListNode's of which only AstNode's
are currently being reused. This fixes `is_reusable_astnode` to also
reused ListNode's if they are the same as before. This bug was found
when the Python grammar was fixed to return the correct body in the
`suite_loop` AST rule.


lib/eco/incparser/incparser.py
```diff
@@ -784,15 +784,28 @@ def interpret_annotation(self, node, production):
...
+
+        if type(old) is not type(new):
             return False
+
         if old.name != new.name:
             return False
-        for key in old.children:
-            if old.children.get(key) is not new.children.get(key):
+
+        if type(old) is ListNode:
```
ltratt, line 795: What are the possible types of `old`. Is it just `ListNode` and `AstNode` or other things too?

ptersilie: It's either `AstNode`, `ListNode`, or terminals and nonterminals.

ltratt: OK, got it.

ltratt (approved):

PR #220, opened 2018-09-19
Fix namebinding

ptersilie:
This PR fixes and improves namebinding across languages, which before needed some language specific hacks to work but now is a bit more language agnostic. It also fixes code completion which broke some time ago.


ptersilie:
The last commit probably needs squashing.

ltratt:
Do you want to do that before I review?

ptersilie:
Squashed!

lib/eco/astanalyser.py
```diff
@@ -260,7 +260,12 @@ def merge_lbox_data(self, node, path):
 
     def convert_uri(self, newkind, prev, path):
         uri = prev
-        if len(prev.path) >= 1:
+        # The top-level in all of our namebinding rules doesn't have a name, so
```
ltratt: First sentence is a bit hard to parse. Rephrase?

lib/eco/astanalyser.py
```diff
@@ -260,7 +260,12 @@ def merge_lbox_data(self, node, path):
...
+        # this is a semi safe way to find the top-level and replace it if there
+        # is one. Alternatively, we could force all namebinding rules to use the
+        # same name, which seems more restrictive.
+        # XXX This doesn't work if a namebinding rule gives the top-level a name.
```
ltratt: This seems to contradict the comment's first sentence. Can we guard against this, or is it impossible, or ...?

ptersilie: We don't need to guard against it, as it doesn't result in an error or something. It just means that the namebinding rules from a language box won't be integrated into the outer language. If we had a documentation we would just say that top-level rules can't have a name. If we want to fix this properly, we have to somehow mark the top-level rule so we can check properly here if a given rule is the top-level or not.

ltratt: Could we guard against it so that if someone *does* specify this, at least they get an error, rather than things just silently not working?

ptersilie: Let me have a look.

ltratt:
OK, please squash.

ltratt:
Ah, let's merge #221 first, then you can squash and rebase against master at the same time.

ptersilie:
Squashed!

ltratt (approved):

PR #225, opened 2018-11-27
Various fixes

ptersilie:
This PR contains the following 3 fixes:

- 553560e fixes the parsetree viewer which previously wouldn't update to the correct parsetree when switching to a different tab within the editor
- 0657a5c is another small change in parsetree viewer that doesn't show BOS/EOS nodes as containing errors. This is only an aesthetic change which avoids confusion when looking at the parse tree.
- f2f5565 fixes error recovery. The fix for the retain bug unfortunately broke out-of-context analysis whose results weren't integrated into the parsetree anymore (even if it succeeded). The remedy for this was to skip empty nonterminals sooner rather than later.

### DIFF
--- a/lib/eco/eco.py
+++ b/lib/eco/eco.py
@@ -1319,6 +1319,9 @@ class Window(QtGui.QMainWindow):
                 self.ui.hud_heat_map_button.setChecked(True)
             else:
                 self.ui.hud_off_button.setChecked(True)
+            tm = ed_tab.editor.tm
+            if self.pgviewer:
+                self.pgviewer.update(tm.main_lbox, tm.version, tm.get_max_version())
         else:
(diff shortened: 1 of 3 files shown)


ptersilie:
The tests currently fail because the sources for the pyqt stuff are down yet again! *sigh* I'm looking into it.

ltratt:
These all look fine to me.

ptersilie:
Tests fixed in #226, so this should pass now.

ptersilie:
Whoops, had to rebase first.

ltratt:
Is this one merge as-is?

ptersilie:
Yes please!

ltratt (approved):

PR #226, opened 2018-11-27
Install PyQt4 dependecy via apt

ptersilie:
Until now we had to build PyQt4 manually since on Travis virtualenv didn't
have access to system packages. This seemed to have changed now so we can go
back to installing these dependecies via apt.


ltratt:
Excellent!

ltratt (approved):

PR #229, opened 2019-02-01
Added SQL grammar adapted from SQLite

ptersilie:
While this grammar is still incomplete it currently parses roughly 90%
of the 37,000 SQLite tests. The remaining 10% are either tests that are
expected to fail (as we currently don't have a way of distinguishing
those) or they require us to solve some ambiguities within the grammar.

Since SQLite uses their own grammar syntax and parser generator
(lemonparser), their grammar has ways to solve ambiguities that are not
available in Yacc. For example, the %fallback directive allows the parser
to fallback to a specified token if all other tokens fail to parse (which
SQLite uses to solve identifier ambiguities). Another directive, %wildcard
allows the parser to parse any token within a certain rule. Since Yacc
doesn't have these directives we need to find other ways to work around
this in the future. But for the time being this grammar is decent enough
to work with, especially compared to the old one.


ltratt:
Agreed, that sounds reasonable. It'll take me a bit to review this because it doesn't display in github...

ltratt:
OK, I've tried a few semi-complex queries out and they seemed to parse as expected, which is good! Three comments:

1) We need to maintain SQLite's licence at the top of the file. Admittedly, this is a bit tricky, because SQLite is public domain. But we should at least say something like "Derived from SQLite, with thanks. SQLite is public domain software: see https://sqlite.org/copyright.html for more details."

2) We need to allow case-insensitive keywords (I was a bit surprised when my first query `select * from t` failed!).

3) There's quite a few random newlines lurking around (e.g. after the "anylist" rule name): let's get rid of 'em.

ptersilie:
I fixed 1 and 2. However, I'm inclined to leave 3 since those newlines make it a bit more obvious that those rules have an empty production as the first rule. This comes down to personal taste tho, so I'm easy to be convinced otherwise.

ltratt:
Is that the order of productions that the original grammar does things in? I must admit, I always make the last production of a rule be the empty production, at which point the problem solves itself.

ptersilie:
Yes, the grammar is automatically generated from SQLite's grammar, so this is how it naturally comes out of the parser I wrote. Tho I reckon I can change the parser to reorder the productions by putting the empty one last.

ltratt:
Up to you. We can keep it as it is.

ptersilie:
I think I rather leave it as is.

ltratt:
OK, please squash.

ltratt:
[Don't forget to rebase as well!]

ptersilie:
Squashed.

ltratt:
Notice "This branch is out-of-date with the base branch". You need to fetch and rebase against origin/master.

ptersilie:
And rebased.

ltratt (approved):

PR #231, opened 2019-02-01
Use gtar if it exists.

ltratt:
The "-O" option is a GNU-ism, which other tars don't share (-O on OpenBSD's tar does something very different). So, if gtar exists, we use that first, falling back on tar otherwise.

I'd be grateful if this could be quickly checked on a desktop Linux box before merging.

### DIFF
--- a/bin/diffeco
+++ b/bin/diffeco
@@ -31,10 +31,18 @@ done
 
 BASEDIR=$(dirname "$0")
 
+# We need a GNU tar, so try gtar first
+which gtar 2>&1 > /dev/null
+if [ $? -eq 0 ]; then
+    TAR=gtar
+else
+    TAR=tar
+fi
+
 if [ -z $file2 ]; then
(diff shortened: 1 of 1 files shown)


bin/diffeco
```diff
@@ -31,10 +31,18 @@ done
 
 BASEDIR=$(dirname "$0")
 
+# We need a GNU tar, so try gtar first
+which gtar 2>1 > /dev/null
```
ptersilie: On Linux this leaves an empty file `1` on the file system.

ptersilie:
Apart from the above comment, works fine.

bin/diffeco
ltratt: Oops, that's a typo. Fixed in the new commit (hopefully). Let me know!

ptersilie:
That fixed it. Please squash!

ltratt:
Squashed.

ptersilie (approved):

PR #232, opened 2019-02-04
Find longest match in Lexer.lex method

ptersilie:
Though Lexer.lex isn't used in Eco (which uses Lexer.get_token_iter),
the method is still useful if we need to lex input strings rather than
nodes. Previously, however, Lexer.lex returned the first matched token
instead of searching for longer matches. This has been fixed now.

I am currently using this in the language-box experiments to extract SQL, PHP, Java, etc. expressions
from source files.

### DIFF
--- a/lib/eco/treelexer/lexer.py
+++ b/lib/eco/treelexer/lexer.py
@@ -472,19 +472,25 @@ class Lexer(object):
         pm = PatternMatcher()
         self.pos = 0
         result = []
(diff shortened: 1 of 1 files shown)


ltratt:
This looks OK, but needs rebasing against master.

ptersilie:
Rebased.

ltratt (approved):

PR #235, opened 2019-04-03
Only show valid languages in submenu

ptersilie:
When opening the language box submenu only show languages that are part
of the composition. Added an option to the settings to revert back to the
old behaviour which shows all languages Eco knows about.

The following screenshot shows the change in action:
submenu

### DIFF
--- a/lib/eco/nodeeditor.py
+++ b/lib/eco/nodeeditor.py
@@ -982,32 +982,63 @@ class NodeEditor(QFrame):
             if lbox:
                 tmp = self.tm.cursor.node
                 self.tm.cursor.node = lbox.prev_term
-
-        # Create actions
-        bf = QFont()
-        bf.setBold(True)
-        valid_langs = []
         lalist = self.tm.getLookaheadList()
-        for l in languages:
-            if "<%s>" % l in lalist:
-                valid_langs.append(l)
-
         if tmp:
             # undo cursor change
             self.tm.cursor.node = tmp
+
+        # Filter out languages that are part of this composition.
+        todo = [self.get_mainlanguage()]
+        comp_langs = []
+        while todo:
+            l = todo.pop()
+            for sl in lang_dict[l].included_langs:
+                if lang_dict[sl] not in comp_langs:
+                    comp_langs.append(lang_dict[sl])
+                    todo.append(sl)
+
+        # Generate list of languages valid at current position
+        valid_langs = []
+        for l in comp_langs:
+            if "<%s>" % l in lalist:
+                valid_langs.append(l)
+        for l in valid_langs:
+            # Don't show languages twice
+            comp_langs.remove(l)
+
+        def create_entry(l, menu):
+            pass
+
+        bf = QFont()
+        bf.setBold(True)
+        # Show valid languages first
         if len(valid_langs) > 0:
-            for l in valid_langs:
+            for l in sorted(valid_langs, key=lambda x: x.name):
                 item = QAction(str(l), menu)
                 item.setData(l)
                 self._set_icon(item, l)
                 item.setFont(bf)
                 menu.addAction(item)
             menu.addSeparator()
-        for l in languages:
+
+        # Then show languages that are part of this composition
+        for l in sorted(comp_langs, key=lambda x: x.name):
             item = QAction(str(l), menu)
             item.setData(l)
             self._set_icon(item, l)
             menu.addAction(item)
+
+        # Finally add another submenu with all remaining languages
+        other_langs = QtGui.QMenu("Other", self)
+        for l in sorted(languages, key=lambda x: x.name):
+            if l not in comp_langs and l not in valid_langs:
+                item = QAction(str(l), other_langs)
+                item.setData(l)
+                self._set_icon(item, l)
+                other_langs.addAction(item)
+        menu.addSeparator()
+        menu.addMenu(other_langs)
+
         return menu
 
     def showSubgrammarMenu(self):


ltratt:
I wonder if we can get the best of both worlds here. I agree that only showing the obviously valid languages in the menu is the right thing, but maybe we can have the bottom menu option be "other languages" and have a sub-menu which shows all the other languages? Then if someone is experimenting (for whatever reason), they can still do so, but we still make the common case fairly uncluttered.

ptersilie:
Sure, that would be possible. Though there is an option to turn on "all languages" in the GUI, which reverts back the old behaviour.

ltratt:
> Though there is an option to turn on "all languages" in the GUI, which reverts back the old behaviour.

In general options aren't a great way of doing this sort of thing: they're hard to understand and hard to remember that they even exist. So if it's plausible to do it another way, I think that would be better.

ptersilie:
Understood. I like your solution, so let's go with that!

ptersilie:
The latest commit implements the above.

lib/eco/nodeeditor.py
```diff
@@ -987,32 +987,31 @@ def createSubgrammarMenu(self, menu, change=False):
...
-                    if lang_dict[sl] not in langs:
-                        langs.append(lang_dict[sl])
-                        todo.append(sl)
-        else:
-            langs = languages
+        # Filter out languages that are part of this composition.
+        todo = [self.get_mainlanguage()]
+        comp_langs = []
+        while todo:
+            l = todo.pop(0)
```
ltratt: Do you need to pop the `0`th element? That will cause quite a bit of shuffling vs. popping the last element.

ptersilie: Not necessarily. I did this mainly to preserve the order in which the languages are composed into the main language. Though I guess it doesn't matter a great deal in which order they show up. On the other hand, this list is so small that this tiny bit of shuffling is not going to impact performance in any way.

ltratt: What order are they displayed in? Alphabetical?

ptersilie: The order is the same as how the user composed them. So:
1. Main language
2. Sub lang 1
3. Sub lang 2
4. Sub lang 1 -> Sub lang
5. Sub lang 2 -> Sub lang

If we change the code to `todo.pop()`, then the order would be slightly different (reversing the sub-sub languages):
1. Main language
2. Sub lang 1
3. Sub lang 2
4. Sub lang 2 -> Sub lang
5. Sub lang 1 -> Sub lang

Admittedly, this is not a big deal.

ltratt: I wonder if we should do it alphabetically?

ptersilie: I think the more immediate compositions should come first, though the order of any language coming after that doesn't really matter. So maybe we just use `todo.pop()`, which still shows the immediate sub languages first.

ltratt: What does "more immediate" mean?

ptersilie: I wasn't sure how to describe this and the first thing that came to mind was "immediate". What I meant was the first level of the composition. So let's say I compose Html + Python + Sql, where Python is composed into HTML, and SQL compose into Python. So in this case I'd like Python to show up before HTML when opening the submenu, as it is comes earlier in "the order of the compositions", i.e. it is less nested, if that makes sense.

ltratt: If you're in HTML and you can only put Python into it, wouldn't SQL be put in the "other" menu?

ptersilie: No, currently the submenu shows all languages that a part of this composition, no matter what language you are in. "Other" shows all other languages that Eco knows about that have nothing todo with the current composition.

ltratt: s/wouldn't/shouldn't/ ;)

ptersilie: I still don't think it should be put there. The "other" menu should only contain languages that have nothing to do with the current composition (and thus should never be used here - which is why I excluded them all together in my initial commit). The main submenu contains all languages that are part of the composition, even if they can't be used in the current language box. This allows us to see at a quick glance, which languages we composed, and since languages valid at the current location are highlighted and move to the top, it don't think this is too confusing.

So basically there are 4 categories:
1. Languages valid at cursor position
2. Languages valid in current box
3. Languages that are part of this composition
4. All other languages Eco knows about

This PR currently merges 2. and 3. together, which IMO is ok.

ltratt: OK, I can live with 2+3 being merged together (presumably there's some sort of spacer between them and things in category 1?). But I think within each category, it's clearest if we present things alphabetically.

ptersilie: Yes, there is a space between 1, 2/3, and 4. I will add a commit sorting the individual categories alphabetically.

ltratt: Thanks!

ptersilie: The latest commit should do the trick.

ltratt:
Please squash.

ptersilie:
Squashed!

ltratt:
I think the commit description needs updating (and fairly heavily expanding). Feel free to force push.

ptersilie:
Description updated!

ltratt:
Spelling: "Prioritise".

ptersilie:
Whoops! Fixed.

ltratt (approved):

PR #237, opened 2019-04-12
Try to continue lexing after error

ptersilie:
When an error occurs during lexing, we abort lexing and leave the tokens
in the tree as is. However, this can lead to tokens containing newlines
to not be split up (which is necessary for Eco to function properly).
We can fix this problem, by ensure that even if an error occurs before
a newly inserted token, at the very least that token is relexed, which
ensures that newly inserted newlines are broken up properly.

### DIFF
--- a/lib/eco/inclexer/inclexer.py
+++ b/lib/eco/inclexer/inclexer.py
@@ -493,7 +493,6 @@ class IncrementalLexerCF(object):
         pos = 0  # read tokens
         read = 0 # generated tokens
         current_node = node
-        sw = StringWrapper(node, startnode)
         next_token = self.lexer.get_token_iter(node).next
 
         combos = []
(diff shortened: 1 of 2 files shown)


lib/eco/inclexer/inclexer.py
```diff
@@ -570,6 +569,14 @@ def relex(self, node):
                 else:
                     # There are no part matches so remark the startnode as error
                     startnode.changed = True
+                if not past_startnode:
+                    # When a lexing error occurs before we reached the newly
+                    # inserted node (startnode) try to continue lex from
```
ltratt: "continue lex*ing*"?

ptersilie: Whoops. Fixed.

ltratt:
Please squash.

ptersilie:
Squashed.

ltratt (approved):

PR #238, opened 2019-04-12
Load external compositions

ptersilie:
REQUIRES REBASE AFTER MERGING #237 

This commit allows us to easily create compositions and load them via
json files into Eco. An example composition of Java and PHP can be found
in compositions/javaphp.json

We might want to use this to create all our compositions and use it to replace all the compositions in `grammars/grammar.py`, as it allows users to easily add compositions without having to edit the Eco source. We can either do this in a separate PR or in this one.

### DIFF
--- a/lib/eco/eco.py
+++ b/lib/eco/eco.py
@@ -867,6 +867,7 @@ class Window(QtGui.QMainWindow):
         parser.add_option("-e", "--export", action="store_true", default=False, help="Fast export files. Usage: --export [SOURCE] [DESTINATION]")
         parser.add_option("-f", "--fullexport", action="store_true", default=False, help="Export files. Usage: --fullexport [SOURCE] [DESTINATION]")
         parser.add_option("-g", "--grammar", action="store_true", default=None, help="Load external grammar. Usage: --grammar [GRAMMARFILE]")
+        parser.add_option("-c", "--composition", action="store_true", default=None, help="Load external composition. Usage: --composition [COMPOSITIONFILE]")
         (options, args) = parser.parse_args()
 
         if options.log.upper() in ["INFO", "WARNING", "ERROR", "DEBUG"]:
@@ -887,6 +888,9 @@ class Window(QtGui.QMainWindow):
             self.cli_export(source, dest, True)
         if options.grammar:
             self.load_external_grammar(args[0])
+            return
+        if options.composition:
+            self.load_composition(args[0])
             return
         if len(args) > 0:
             for f in args:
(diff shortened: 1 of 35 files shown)


lib/eco/eco.py
```diff
@@ -1142,6 +1146,24 @@ def load_external_grammar(self, filename):
         etab.editor.customContextMenuRequested.connect(self.contextMenu)
         self.toggle_menu(True)
 
+    def load_composition(self, filename):
+        import json
+        from grammars import grammars
+        with open(filename) as f:
+            comp = json.load(f)
+            print(comp)
```
ltratt: Debugging statement?

ltratt:
I'd suggest that we might as well move everything over to use this style in this PR, then it's done and dusted!

ptersilie:
The latest commit converts all previous languages and compositions into the new JSON format. They can now be found in `grammars/include/`.

There were a few pitfalls with circular references and accidentally overwriting existing languages, which are caught in https://github.com/softdevteam/eco/pull/238/commits/607ab5b91b59241aee1243f947ad4fa76ae7efe3#diff-dd418cfeb38ea6e88b48934b4c685b5fR212 and https://github.com/softdevteam/eco/pull/238/commits/607ab5b91b59241aee1243f947ad4fa76ae7efe3#diff-dd418cfeb38ea6e88b48934b4c685b5fR212.

For example, https://github.com/softdevteam/eco/pull/238/commits/607ab5b91b59241aee1243f947ad4fa76ae7efe3#diff-e5434c64a011db603a0f9e86f4d198dc uses circular references, since `Python + PHP` references back to `PHP + Python`. This example also shows how we can reuse existing languages (e.g.~when only a `name` and `location` is supplied, the composition tries to use the existing languages if it exists; otherwise an error is shown).

Similarly, if one tries to redefine an existing language within the `composition` field, an error is shown and the editor exits.

ltratt:
This looks good to me. My only (minor) question is whether we should take this opportunity to think/rethink about some of the composition names. Some of the more complex ones aren't very intuitive (e.g. "Java + SQL ref. Java"). I'm not honestly sure if we can do much about that, but it might be worth having a quick think about whether there's some sort of naming scheme that might help (e.g. would "Java(Sql(Java expressions))" be better?).

ptersilie:
I've already renamed some of the grammars. To be fair, the particular one you mentioned is only needed for tests, so I have now move it there, so it's not visible to the user anymore.

ltratt:
I am still wondering if the `inner(outer)` naming is clearer than `inner + outer` though. Any thoughts?

ptersilie:
Dunno, looks a bit weird and won't work for circular stuff, e.g.
`PHP(Python(PHP(Python(PHP(Python...`. I agree that `PHP + Python` doesn't really tell us that Python can reference PHP, but I don't see a solution for this with either formatting.

ltratt:
Point taken. Maybe have a think about it for the future in case you have a brainwave!

ltratt:
Please squash.

ptersilie:
Sure, I'll keep it in the back of my mind and make a PR if I come up with a better solution. Squashed!

ltratt:
You need to rebase too.

ptersilie:
Rebased.

ltratt (approved):

PR #241, opened 2019-04-24
Fix inclexer

ptersilie:
REQUIRES MERGING OF #240 PLUS REBASE!

The tree lexer (which works on a parse tree instead of a steam of
tokens) needs to remember which nodes it has seen during lexing so that
the incremental lexer can then merge new tokens with these nodes.

When the lexer sees a node and it is not already in the list of seen
nodes, it is added. Unfortunately, the check for this previously used
equality to determine if the node has been seen before, which leads to
errors when two or more equivalent nodes are processed during relexing.
This commit updates the existence check to use identity instead, which
fixes the problem.

### DIFF
--- a/lib/eco/test/test_eco.py
+++ b/lib/eco/test/test_eco.py
@@ -1520,6 +1520,24 @@ class Test_JavaBugs(Test_Java):
         self.treemanager.key_normal("'")
         assert self.treemanager.cursor.node.symbol.name == "'1 '"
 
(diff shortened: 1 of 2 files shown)


ltratt:
This looks fine (though I had to read it twice to work out what's going on -- it's subtle!). Please rebase.

ptersilie:
Rebased.

ltratt (approved):

PR #243, opened 2019-05-16
Add another heuristic for automatic language boxes

ptersilie:
The current heuristic finds automatic language box candidates by
searching for valid locations on the current parse stack where a
language box could be inserted. The problem with this is that some
grammars apply reductions too early which means we are missing out on
many valid locations.

This commit adds a second heuristic which traverses all parents of the
error node in the previous parse tree and attempts to insert candidates
just before them, thus finding many locations that a stack traversal
wouldn't.

However, since the new heuristic depends on having a previous parse
tree, it doesn't work when the currently typed program has never been
valid, e.g. the user has just started writing code. We thus run both
heuristic in conjunction. We first run heuristic #2 (traversing parents)
and only if it fails to find candidates, then we run heuristic #1
(traversing the parse stack). This gives us the best of both worlds and
increases the accuracy of automatically inserted language boxes.

### DIFF
--- a/lib/eco/autolboxdetector.py
+++ b/lib/eco/autolboxdetector.py
@@ -42,24 +42,95 @@ class NewAutoLboxDetector(object):
             node = node.parent
         return node.right_sibling()
 
-    def find_terminal(self, node):
+    def find_terminal(self, node, version=None):
         startnode = node
         while type(node) is not BOS:
-            if node.children:
-                node = node.children[-1]
+            if node.get_attr("children", version):
+                node = node.get_attr("children", version)[-1]
             elif type(node.symbol) is Terminal:
                 break
             else:
                 if node.new:
                     return
-                while not node.left_sibling():
-                    node = node.parent
+                while not node.get_attr("left", version):
+                    node = node.get_attr("parent", version)
                     if node is startnode:
                         return None
-                node = node.left_sibling()
+                node = node.get_attr("left", version)
         return node.next_term
 
     def detect_lbox(self, errornode):
+        # Try heuristic #2 first; if it fails to find an automatic language
+        # box, try heuristic #1
+        valid = self.detect_lbox_h2(errornode)
+        if not valid:
+            valid = self.detect_lbox_h1(errornode)
+
+        if errornode.autobox is False:
+            # XXX Currently, we don't suggest any language boxes for an error
+            # if the user had to revert an automatically inserted box. However,
+            # we might want to consider showing suggestions if the suggested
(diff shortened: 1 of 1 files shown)


lib/eco/autolboxdetector.py
```diff
@@ -42,24 +42,98 @@ def pop_lookahead(self, node):
...
+        if not valid:
+            valid = self.detect_lbox_h1(errornode)
+
+        if errornode.autobox is False:
+            # XXX Currently, we don't suggest any language boxes for an error,
+            # if the user had to revert an automatically inserted box. However,
+            # we might want to consider showing suggestions if the suggested box
+            # differs from the one the user reverted
+            return False # don't use this node for autoboxes anymore
+        if valid:
```
ltratt: What are the possible values of `valid` here? `True/False/None`...?

lib/eco/autolboxdetector.py
```diff
@@ -42,24 +42,98 @@ def pop_lookahead(self, node):
...
+    def parse_after_lbox_h2(self, lbox, end, parent):
+        root = parent.get_root()
+        p, l = lang_dict[root.name].load()
+        ir = IncrementalRecognizer(p.syntaxtable, l.lexer, root.name, None)
+        if root is not parent:
+            # if the parent is already the root we don't need to preparse
+            # anything
+            ir.preparse(root, parent)
+        # try parsing lbox + one more non-ws terminal
+        if ir.parse_single(TextNode(lbox)):
```
ltratt: You could simplify these lines to `return ir.parse_single(TextNode(lbox)) and ir.parse_after(end.next_term)`.

lib/eco/autolboxdetector.py
ptersilie: Valid is a list which contains all candidates found by the heuristics. Should I change this to `if len(valid) > 0`?

ptersilie: @ltratt Ping.

ltratt: Yes, definitely. [Boolean checks in Python are a disaster: the empty list comparing equal to `False` is a terrible decision!]

ptersilie:
Addressed comments.

ltratt:
Please squash.

ptersilie:
Squashed!

ltratt:
You need to rebase against master too.

ptersilie:
Rebased.

ltratt (approved):

PR #244, opened 2019-05-16
Improve automatic language boxes

ptersilie:
**Requires merging #243 first.**

This PR adds various improvements and additions to automatic language boxes:
- e5653c8: Allows language boxes to expand if tokens following the box can be moved inside the box. This commit also improves loose/fixed boxes by fixating automatically inserted (loose) language boxes when the user manually moves the cursor out of them.
- e4f5bc3: Extends e5653c8. If there are multiple expansion possibilities, they are shown via the lightbulb icon on the left.
- 4df519f: Added long overdue option to remove language boxes and interpret their content in the outer language.
- 8f39a27: Use error recovery during the automatic removal check.
- 78c5943: Allows language boxes to shrink again if they contain a parse error.
- 830798d: Fixes e5653c8 to only fixate language boxes if they were manually chosen by the user.
- 17a426b: Fix: Makes sure that automatically inserted language boxes are reparsed in a single step (previously we would reparse after deletion of the old nodes and then again after inserting the language box).
- 437fd5f: Improves e5653c8. Only expand language boxes if this doesn't introducing errors immediately after the box.

### DIFF
--- a/lib/eco/autolboxdetector.py
+++ b/lib/eco/autolboxdetector.py
@@ -21,8 +21,10 @@ class NewAutoLboxDetector(object):
     occurs. Similar to error recovery it uses the current parse stack to
     determine a location where a language box can be inserted and then tries to
     wrap the error into that box."""
-    def __init__(self, origparser):
+    def __init__(self, origparser, origlexer):
         self.op = origparser
+        self.ol = origlexer
+        self.olang = None
         self.langs = {}
         self.mode_limit_tokens_new = False
 
@@ -31,6 +33,7 @@ class NewAutoLboxDetector(object):
             return
 
         main = lang_dict[langname]
+        self.olang = langname
         self.mode_limit_tokens_new = main.auto_limit_new
 
         # preload nested languages
@@ -218,6 +221,14 @@ class NewAutoLboxDetector(object):
             if type(element) is Accept:
                 return True
             return False
+
+    def check_remove_lbox(self, lbox):
+        r = IncrementalRecognizer(self.op.syntaxtable, self.ol.lexer, self.olang, None)
+        stack = [s.state for s in self.op.stack]
+        r.state = stack
+        result = r.parse(lbox.symbol.ast.children[0].next_term, lbox.next_term, self.op.last_status)
+        if result:
+            lbox.tbd = "remove"
 
 from inclexer.inclexer import StringWrapper
 from treelexer.lexer import LexingError
--- a/lib/eco/autolboxdetector.py
+++ b/lib/eco/autolboxdetector.py
@@ -343,6 +343,16 @@ class Recognizer(object):
             else:
                 return False
 
+    def parse_lex_single(self, node):
+        self.tokeniter = self.lexer.get_token_iter(node).next
+        token = self.next_token()
+        while True:
+            if not self.temp_parse(self.state, token):
+                return False
+            token = self.next_token()
+            if self.last_read is not node or type(token) is FinishSymbol:
+                return True
+
 class RecognizerIndent(Recognizer):
 
     def __init__(self, syntaxtable, lexer, lang, outer):
@@ -499,6 +509,9 @@ class IncrementalRecognizer(Recognizer):
             if self.parse_after(follow):
                 return True
         return False
+
+    def parse_lex_single(self, node):
+        return Recognizer.parse_lex_single(self, node)
 
     def parse_until(self, start, end):
         node = start.next_term
--- a/lib/eco/autolboxdetector.py
+++ b/lib/eco/autolboxdetector.py
@@ -97,17 +97,33 @@ class NewAutoLboxDetector(object):
                     left = parent.get_attr("children", pv)[0] # bos
                 else:
                     left = parent.get_attr("left", pv)
+                while left and type(left.symbol) is Nonterminal and len(left.get_attr("children", pv)) == 0:
+                    # If left is an empty nonterminal, keep going left until we
+                    # find a non-empty nonterminal or a terminal
+                    left = left.get_attr("left", pv)
                 if left:
                     state = left.state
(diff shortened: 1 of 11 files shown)


ptersilie:
Addressed comments.

ptersilie:
Added a fix for the automatic language box expansion, which wasn't working properly for some examples.

lib/eco/treemanager.py
```diff
@@ -2191,16 +2191,17 @@ def lbox_expand_test(self, lbox):
...
-        start = lbox
         for e in r.possible_ends:
             if e.lookup == "<ws>" or e.lookup == "<return>":
                 continue
-            if r.parse_until(start, e):
+            tmp = list(r2.state)
+            if r2.parse_after(e.next_term):
                 filtered.append(e)
-            start = e
+            r2.state = tmp
```
ltratt, line 2206: Why are we (in effect) doing `r2.state = list(r2.state)`? What's the `list` call for?

ptersilie: `state` is the (admittedly poorly named) parsing stack of the recogniser. After initialising the recogniser with `preparse`, we then want to try and check if each element in `possible_end` can be parsed in that state, so we have to reset the parse stack back after every iteration.

ltratt: Yes, but it's the `list` bit that I'm confused by.

ptersilie: It just creates a copy of the parse stack which we can then later apply to r2, to reset the parse stack back to what it was before.

ptersilie: I can also use `r2.state[:]` if you prefer that.

ltratt: Oh, I see, you're doing a shallow copy! Ah. Yes, `[:]` will make this clearer. I thought you were converting from a non-list to a list!

ptersilie: Fixed in the latest commit.

ltratt:
Tests failed?

ptersilie:
Weird, looks fine on travis. I'll just restart.

ptersilie:
Nevermind, found it now. Will fix.

ptersilie:
Okay, this is taking a bit longer than I expected. I'll let you know when this is fixed.

ltratt:
OK.

ptersilie:
This should fix things. I also had to rebase as the error would only appear when merged into master.

lib/eco/autolboxdetector.py
```diff
@@ -97,16 +97,22 @@ def detect_lbox_h2(self, errornode):
...
                         term = self.find_terminal(left, pv)
                         if type(term) is EOS:
+                            parent = parent.get_attr("parent", pv)
                             continue
                         while term and term.lookup in ws:
                             # skip whitespace
                             term = term.next_term
+                        element = self.op.syntaxtable.lookup(term.state, lbox)
+                        if type(element) not in [Reduce, Shift]:
+                            # can we still shift lbox?
```
ltratt: What does the comment mean?

ptersilie: We have already checked that the lbox can be shifted after the nonterminal `left`. This is a sanity check, that tests if lbox can still be shifted after `term` which is the most-right terminal in `left`. This should almost always be true, so this double checks that this is really the case. I believe there can be cases, where the terminal is an error node, in which case this second check will fail, so that we don't insert a box which then turns out to be invalid when reparsed.

ltratt: Do we have a test case for this? If not, can we maybe expand the comment along the lines of what you've just written.

ptersilie: I think I need to add some tests for this and the other features, as I just found another bug with the new stuff.

ptersilie: Although in this case, I don't currently have a test, though I might be able to come up with one.

ltratt: Let's give it a go.

ptersilie:
I've just added a couple of tests for the new improvements, and also clarified the comment for which I couldn't find a test.

lib/eco/test/test_eco.py
```diff
@@ -5247,3 +5247,44 @@ def test_java_lua_dont_remove_explicit_lboxes(self):
...
+        treemanager = TreeManager()
+        treemanager.option_autolbox_insert = True
+        treemanager.add_parser(parser, lexer, "")
+        p = """class X {
+    int x = 1 == 2;
+}"""
+        for c in p:
+            treemanager.key_normal(c)
+        assert len(treemanager.parsers) == 1
+        assert parser.last_status == True
```
ltratt, line 5266: You don't need `== True` yet.

ltratt: s/yet/here/

lib/eco/test/test_eco.py
```diff
@@ -5247,3 +5247,44 @@ def test_java_lua_dont_remove_explicit_lboxes(self):
...
+
+        treemanager.key_end()
+        treemanager.key_cursors(LEFT)
+        treemanager.key_backspace()
+        treemanager.key_backspace()
+        treemanager.key_backspace()
+        treemanager.key_backspace()
+
+        assert len(treemanager.parsers) == 2 # box now has been expanded
+        assert parser.last_status == True
```
ltratt, line 5292: Ditto.

ltratt:
Where are with this one?

ptersilie:
I'm almost done going through the logs and fixing bugs / adding features when I encounter problems.

ptersilie:
Alright, I think we can get this squashed and merged now.

ltratt:
OK I checked the new commits and they look OK. Please squash.

ptersilie:
Squashed!

ltratt (approved):

PR #245, opened 2019-05-31
Fix CHARACTER_LITERAL in Java grammar

ptersilie:
This updates the lexing rule for character literals in Java according to spec. Allowed between single quotes are:

- any printable character other than a backslash or '
- Escaped characters: `\b \t \n \f \r \" \' \\`
- Escaped octal number between 000 and 377 (e.g. `\123`)
- Esacped Unicode, e.g. `\u012A`

To view the diff of these changes use `./bin/diffeco lib/eco/grammars/java15.eco -c master`.


ltratt:
Do we need to make an update to https://github.com/softdevteam/grammars/blob/master/java7/java.l too based on this?

ptersilie:
I would think so. Remind me what `?:\\\\` does again?

ltratt:
We use https://docs.rs/regex/1.1.6/regex/ syntax in that repo.

ltratt:
So `(?:\\\\|blah)` means "match `\\` or `blah` but don't make a group for that".

ptersilie:
I see. Should I update that repo as well or do you want to do that?

ltratt:
I will happily accept a PR to the grammars repo :)

ltratt:
Is this ready for squashing?

ptersilie:
Squashed!

ltratt (approved):

PR #246, opened 2019-06-03
Line heuristic

ptersilie:
This PR adds a line-based heuristic for automatic language box detection. In order to find locations at which language boxes can be inserted, this heuristic simply scans all tokens up to the beginning of the line. While this heuristic performs really well given some preliminary experiments, there are still edge cases where the other heuristics are needed, for example in Java+PHP:
```
class x{
    function x(){
        $x = 1; // Error happens in this line
    }
}
```

This PR also moves the configuration of which heuristic is to be used to a new config module `config.py` to which we will move over all other editor configuration options in a separate PR.

Update: This PR now also runs all heuristics and then ranks their results by parsing distance, discarding non-maximal results.


ptersilie:
This PR is currently on hold, as the line-based approach doesn't work in the following composition of Java and PHP:
```Java
class X {
    int x = !e($x) ? $y : $z == 2; // this used to be 1 == 2
}
```
The line-based approach first creates a box around `!e($x)` and then around `$y == 2`. But when we continue typing `: $z`, the first box doesn't get extended. For this to work we need to first add another tweak that allows the automatic lbox detector to break down the second language box when trying to extend the first one.

ptersilie:
I have updated this PR and it is now ready for review.

ltratt:
Are the comments at the top of the PR still correct / relevant?

ptersilie:
To keep the PR small I have decided to ignore the broken test this causes and do the fix in a separate PR.

ptersilie:
> Are the comments at the top of the PR still correct / relevant?

I have added an update to the PR description.

lib/eco/autolboxdetector.py
```diff
@@ -82,6 +82,31 @@ def detect_lbox(self, errornode):
...
+                    r = self.langs[sub]
+                    r.mode_limit_tokens_new = self.mode_limit_tokens_new
+                    start = node.next_term
+                    result = r.parse(start)
+                    if r.possible_ends:
+                        for e in r.possible_ends:
+                            if e.lookup == "<ws>" or e.lookup == "<return>":
+                                continue
+                            if (self.contains_errornode(start, e, errornode) \
+                                    and self.parse_after_lbox_h2(lbox, e, start, pv)):
```
ltratt: This formatting is very hard to read. Maybe too much indentation or ...?

lib/eco/autolboxdetector.py
```diff
@@ -63,11 +64,16 @@ def find_terminal(self, node, version=None):
...
-        valid = self.detect_lbox_h2(errornode)
-        if not valid:
-            valid = self.detect_lbox_h1(errornode)
+        # Try history based heuristic first; if it fails to find an automatic
+        # language box, try the stack based one; if that fails, try the
+        # line-based heuristic.
+        valid = []
+        if config.AUTOLBOX_HEURISTIC_HIST:
+            valid = self.heuristic_history(errornode)
+        if not valid and config.AUTOLBOX_HEURISTIC_STACK:
```
ltratt: An obvious question is what happens if we remove the `not valid and` here (and remove duplicates)?

ltratt: Ah, OK, this is addressed in the following commit.

lib/eco/autolboxdetector.py
```diff
@@ -68,12 +68,26 @@ def detect_lbox(self, errornode):
...
-            valid = self.heuristic_line(errornode)
+            valid.extend(self.heuristic_history(errornode))
+        if config.AUTOLBOX_HEURISTIC_STACK:
+            valid.extend(self.heuristic_stack(errornode))
+        if config.AUTOLBOX_HEURISTIC_LINE:
+            valid.extend(self.heuristic_line(errornode))
+
+        maxdistance = {}
+        # find max distance by startnode
+        for start, end, lang, dist in valid:
```
ltratt: `for start, _, _, dist in valid`?

lib/eco/treemanager.py
```diff
@@ -2218,7 +2218,8 @@ def lbox_expand_test(self, lbox):
         r2.preparse(outer_root, lbox)
         r2.parse_after(lbox)
         filtered = []
-        for e in r.possible_ends:
+        # XXX filter by distance
```
ltratt: `XXX`?

lib/eco/autolboxdetector.py
ptersilie: I'm not sure how I can improve this.

ltratt: A nested `if` would probably be less worse. I had to read this 3 or 4 times to work out what was going on.

lib/eco/treemanager.py
ptersilie: This was a reminder that I need to filter the possible results for expanding a language box by removing all non-maximal results (similar to the creating of language boxes). However, now that I have implemented this I am unsure if that is what we want. Because it means that we will always just expand to the longest match and not show other possible expansion options. Thoughts?

ltratt: We might be able to determine this experimentally: if you run the experiment with both, does the number of acceptable outcomes differ in a meaningful way?

ptersilie: I can try. We will likely see less "No insertion (Multi)", since we will always pick the longest match. This won't tell us though if the other options are something the user wanted to do since they may not always want to expand to the longest match.

ltratt: Your call. I don't really know what's best, either.

ptersilie: I'm always a bit reluctant removing options from the user, so I think I'm just gonna leave it as is for now and remove this comment. I've saved the filtering code, so we can bring it back if we decide it's the better option.

ltratt: I think we should run the experiment, because it's an almost zero-cost thing to do. If, as is likely, it doesn't tell us very much then we can keep things as-is. Otherwise, AFAICS, we'll never have any other data to guide our decision...

ptersilie:
See if these make things better!

lib/eco/treemanager.py
ptersilie: Okay, I'll start a run and see what happens.

lib/eco/autolboxdetector.py
```diff
@@ -78,7 +78,7 @@ def detect_lbox(self, errornode):
 
         maxdistance = {}
         # find max distance by startnode
-        for start, end, lang, dist in valid:
+        for start, _, _, dist in valid:
```
ltratt, line 81: ISTR there was another one of these a few lines below that's probably worth the same treatment.

ptersilie: I think that one uses all elements of the tuple.

ltratt: Ah OK.

ltratt:
If the test pass, please squash.

lib/eco/treemanager.py
ptersilie: Looking at the results, this change didn't really seem to improve our overall results, so I'm just leaving things as is, as I believe multiple options are in this case better than no options.

ptersilie:
Squashed!

ltratt (approved):

PR #253, opened 2019-07-04
Grammar fixes

ptersilie:
This PR fixes some problems with the PHP and SQlite grammars.

sqlite.eco:
```diff
@@ -672,6 +672,7 @@ xfullname ::=
 %right "BITNOT"
 %nonassoc "ON"
 %%
+<ws>:"/\*.*?\*/"
 ID:"OFF|BLOB"
 ABORT:"[Aa][Bb][Oo][Rr][Tt]"
 ACTION:"[Aa][Cc][Tt][Ii][Oo][Nn]"
@@ -842,6 +843,5 @@ ID:"`([^`]|``)+`"
 ID:"\[[^\]]+\]"
 ID:"[a-zA-Z_][a-zA-Z0-9_]*"
 <ws>:"\-\-[^\\r]*"
-<ws>:"/\*.*?\*/"
 <ws>:"[ \\t]+"
 <return>:"[\\n\\r]"
```

php.eco:
```diff
@@ -16,7 +16,7 @@ top_statement ::=
                statement
        |       function_declaration_statement
        |       class_declaration_statement
-       |       "T_HALT_COMPILER" "(" ")" ";"
+       |       "T_HALT_COMPILER" "LP" "RP" ";"
        |       "T_NAMESPACE" namespace_name ";"
        |       "T_NAMESPACE" namespace_name "{"
                top_statement_list "}"
@@ -79,7 +79,7 @@ inner_statement ::=
                statement
        |       function_declaration_statement
        |       class_declaration_statement
-       |       "T_HALT_COMPILER" "(" ")" ";"
+       |       "T_HALT_COMPILER" "LP" "RP" ";"
 ;
 
 
@@ -95,13 +95,13 @@ unticked_statement ::=
        |       "T_WHILE"  parenthesis_expr  while_statement
        |       "T_DO"  statement "T_WHILE"  parenthesis_expr ";"
        |       "T_FOR"
-                       "("
+                       "LP"
                                for_expr
                        ";"
                                for_expr
                        ";"
                                for_expr
-                       ")"
+                       "RP"
                        for_statement
        |       "T_SWITCH" parenthesis_expr      switch_case_list
        |       "T_BREAK" ";"
@@ -117,16 +117,16 @@ unticked_statement ::=
        |       "T_ECHO" echo_expr_list ";"
        |       "T_INLINE_HTML"
        |       expr ";"
-       |       "T_UNSET" "(" unset_variables ")" ";"
-       |       "T_FOREACH" "(" variable "T_AS"
+       |       "T_UNSET" "LP" unset_variables "RP" ";"
+       |       "T_FOREACH" "LP" variable "T_AS"
 
-               foreach_variable foreach_optional_arg ")"
+               foreach_variable foreach_optional_arg "RP"
                foreach_statement
-       |       "T_FOREACH" "(" expr_without_variable "T_AS"
+       |       "T_FOREACH" "LP" expr_without_variable "T_AS"
 
-               foreach_variable foreach_optional_arg ")"
+               foreach_variable foreach_optional_arg "RP"
                foreach_statement
-       |       "T_DECLARE"  "(" declare_list ")" declare_statement
+       |       "T_DECLARE"  "LP" declare_list "RP" declare_statement
        |       ";"
        |       "T_TRY"  "{" inner_statement_list "}"
                catch_statement
@@ -137,9 +137,9 @@ unticked_statement ::=
 
 catch_statement ::=
 
-       |       "T_CATCH" "("
+       |       "T_CATCH" "LP"
                fully_qualified_class_name
-               "T_VARIABLE" ")"
+               "T_VARIABLE" "RP"
                "{" inner_statement_list "}"
                additional_catches;
 
@@ -159,7 +159,7 @@ non_empty_additional_catches ::=
 ;
 
 additional_catch ::=
-       "T_CATCH" "(" fully_qualified_class_name  "T_VARIABLE" ")"  "{" inner_statement_list "}"
+       "T_CATCH" "LP" fully_qualified_class_name  "T_VARIABLE" "RP"  "{" inner_statement_list "}"
 ;
 
 unset_variables ::=
@@ -191,7 +191,7 @@ is_variadic ::=
 
 unticked_function_declaration_statement ::=
                function is_reference "T_STRING"
-               "(" parameter_list ")"
+               "LP" parameter_list "RP"
                "{" inner_statement_list "}" {FunctionDefinition(name=#2, params=#4, body=#7)}
 ;
 
@@ -250,7 +250,7 @@ foreach_optional_arg ::=
 foreach_variable ::=
                variable
        |       "&" variable
-       |       "T_LIST" "("  assignment_list ")"
+       |       "T_LIST" "LP"  assignment_list "RP"
 ;
 
 for_statement ::=
@@ -357,9 +357,9 @@ optional_class_type ::=
 
 
 function_call_parameter_list ::=
-               "(" ")"
-       |       "(" non_empty_function_call_parameter_list ")" {#1}
-       |       "(" yield_expr ")"
+               "LP" "RP"
+       |       "LP" non_empty_function_call_parameter_list "RP" {#1}
+       |       "LP" yield_expr "RP"
 ;
 
 
@@ -407,7 +407,7 @@ class_statement ::=
                variable_modifiers  class_variable_declaration ";"
        |       class_constant_declaration ";"
        |       trait_use_statement
-       |       method_modifiers function is_reference "T_STRING"       "(" parameter_list ")" method_body {FunctionDefinition(name=#3, params=#5, body=#7)}
+       |       method_modifiers function is_reference "T_STRING"       "LP" parameter_list "RP" method_body {FunctionDefinition(name=#3, params=#5, body=#7)}
 ;
 
 trait_use_statement ::=
@@ -551,7 +551,7 @@ new_expr ::=
 ;
 
 expr_without_variable ::=
-               "T_LIST" "("  assignment_list ")" "=" expr
+               "T_LIST" "LP"  assignment_list "RP" "=" expr
        |       variable "=" expr {VariableAssignment(name=#0, value=#2)}
        |       variable "=" "&" variable
        |       variable "=" "&" "T_NEW" class_name_reference  ctor_arguments
@@ -604,7 +604,7 @@ expr_without_variable ::=
        |       expr "T_INSTANCEOF" class_name_reference
        |       parenthesis_expr
        |       new_expr
-       |       "(" new_expr ")"  instance_call
+       |       "LP" new_expr "RP"  instance_call
        |       expr "?"
                expr ":"
                expr
@@ -627,10 +627,10 @@ expr_without_variable ::=
        |       "T_PRINT" expr
        |       "T_YIELD"
        |       function is_reference
-               "(" parameter_list ")" lexical_vars
+               "LP" parameter_list "RP" lexical_vars
                "{" inner_statement_list "}"
        |       "T_STATIC" function is_reference
-               "(" parameter_list ")" lexical_vars
+               "LP" parameter_list "RP" lexical_vars
                "{" inner_statement_list "}"
 ;
 
@@ -647,7 +647,7 @@ combined_scalar_offset ::=
     | string "[" dim_offset "]";
 
 combined_scalar ::=
-      "T_ARRAY" "(" array_pair_list ")"
+      "T_ARRAY" "LP" array_pair_list "RP"
     | "[" array_pair_list "]";
 
 function ::=
@@ -656,7 +656,7 @@ function ::=
 
 lexical_vars ::=
 
-       |       "T_USE" "(" lexical_var_list ")"
+       |       "T_USE" "LP" lexical_var_list "RP"
 ;
 
 lexical_var_list ::=
@@ -726,7 +726,7 @@ dynamic_class_name_variable_property ::=
 
 exit_expr ::=
 
-       |       "(" ")"
+       |       "LP" "RP"
        |       parenthesis_expr
 ;
 
@@ -772,7 +772,7 @@ static_scalar_value ::=
        |       namespace_name
        |       "T_NAMESPACE" "T_NS_SEPARATOR" namespace_name
     |  "T_NS_SEPARATOR" namespace_name
-       |       "T_ARRAY" "(" static_array_pair_list ")"
+       |       "T_ARRAY" "LP" static_array_pair_list "RP"
        |       "[" static_array_pair_list "]"
        |       static_class_constant
        |       "T_CLASS_C"
@@ -812,7 +812,7 @@ static_operation ::=
        |       static_scalar_value "?" static_scalar_value ":" static_scalar_value
        |       "+" static_scalar_value
        |       "-" static_scalar_value
-       |       "(" static_scalar_value ")"
+       |       "LP" static_scalar_value "RP"
 ;
 
 
@@ -853,8 +853,8 @@ expr ::=
 ;
 
 parenthesis_expr ::=
-               "(" expr ")"
-       |       "(" yield_expr ")"
+               "LP" expr "RP"
+       |       "LP" yield_expr "RP"
 ;
 
 
@@ -986,7 +986,7 @@ assignment_list ::=
 
 assignment_list_element ::=
                variable
-       |       "T_LIST" "("  assignment_list ")"
+       |       "T_LIST" "LP"  assignment_list "RP"
        |
 ;
 
@@ -1034,12 +1034,12 @@ encaps_var_offset ::=
 
 
 internal_functions_in_yacc ::=
-               "T_ISSET" "(" isset_variables ")"
-       |       "T_EMPTY" "(" variable ")"
-       |       "T_EMPTY" "(" expr_without_variable ")"
+               "T_ISSET" "LP" isset_variables "RP"
+       |       "T_EMPTY" "LP" variable "RP"
+       |       "T_EMPTY" "LP" expr_without_variable "RP"
        |       "T_INCLUDE" expr
        |       "T_INCLUDE_ONCE" expr
-       |       "T_EVAL" "(" expr ")"
+       |       "T_EVAL" "LP" expr "RP"
        |       "T_REQUIRE" expr
        |       "T_REQUIRE_ONCE" expr
 ;
@@ -1156,6 +1156,8 @@ T_ARRAY_CAST:"\(array\)"
 T_OBJECT_CAST:"\(object\)"
 T_BOOL_CAST:"\(bool\)"
 T_UNSET_CAST:"\(unset\)"
+LP:"\("
+RP:"\)"
 T_NEW:"new"
 T_CLONE:"clone"
 T_EXIT:"exit"
```


ltratt:
Is it just me or is `ws` still defined twice after this commit?

```
<ws>:"/\*.*?\*/"
<ws>:"\-\-[^\\r]*"
```

ptersilie:
Yes, I just lex comments into `<ws>` tokens, because that's essentially what they are. Though now that I think about it, I remember that in the other grammars we use a special `<comment>` token which similarly to the `<ws>` token gets attached to the `WS` nonterminal which is inserted after every token. Not sure which approach is better to be honest, but I guess we should keep things the same in every grammar. Which approach should we use?

ltratt:
I mean: there are two lexing rules both called `ws`. Doesn't that cause some sort of problem?

ptersilie:
No, why should it? It just means that different regex's can be matched to the same token name. It's essentially the same as writing a single regex `regex1|regex2`.

ltratt:
Ah! I guess that's because Eco doesn't map names to integer (doing the same thing in grmtools would cause horrible problems, so it's an error). OK.

ltratt:
> Though now that I think about it, I remember that in the other grammars we use a special <comment> token which similarly to the <ws> token gets attached to the WS nonterminal which is inserted after every token. Not sure which approach is better to be honest, but I guess we should keep things the same in every grammar. Which approach should we use?

I guess consistency is a virtue in this case so we should tweak it?

ptersilie:
> Ah! I guess that's because Eco doesn't map names to integer (doing the same thing in grmtools would cause horrible problems, so it's an error). OK.

No, the token name has no other use than to map it to the grammar. Priorities are extracted from their index in the token list.

> I guess consistency is a virtue in this case so we should tweak it?

I agree, I'm just not sure which approach is better, since the `<comment>` one comes with some boilerplate, but on the other hand is more flexible. E.g. this is how it looks like in Java:
```
comment ::= mlcomment | slcomment;
mlcomment ::= "<mlcomment>";
slcomment ::= "<slcomment>";

%%
%implicit_ws=true
%comment_rule=comment
%nowhitespace=mlcomment,slcomment
%%
<slcomment>:"//[^\\r]*"
<mlcomment>:"/\*.*?\*/"
```

This means users can define their own `comment` rule and their own different comment tokens, but then also need to tell the grammar what the name of their comment rule is, so the grammar can attach it to the `WS` nonterminal.

ltratt:
You should make the call with the "comment" thing: I'm happy either way.

ptersilie:
I definitely want to keep the comment-rule approach in general, since it gives the user more flexibility. Maybe we should leave the other approach in as well, so that we have an example for both approaches?

ltratt:
I don't think we need an example of both if one is better than the other.

ptersilie:
I'm not sure one is better than the other. One gives you more flexibility, and the other is shorter. It really depends on the use case.

ltratt:
OK, my efforts to tempt you off the fence have failed, so I'm going to take the hint ;) Do you want to squash before I merge?

ptersilie:
Hah, well I'm really not sure what the best way is so I was hoping you have a preference. ;-)

No need to squash.

ltratt (approved):

PR #255, opened 2019-07-25
Fix a couple of bugs

ptersilie:
A small collection of bugfixes that I cherry-picked out of the line-heuristic branch, in preparation of another PR.

### DIFF
--- a/lib/eco/treemanager.py
+++ b/lib/eco/treemanager.py
@@ -1569,9 +1569,20 @@ class TreeManager(object):
         l = []
         n = newend
         while type(n) is not EOS:
-            l.append(n.symbol.name)
+            if isinstance(n, MultiTextNode):
+                for c in list(n.children):
+                    l.append(c.symbol.name)
+                    c.remove()
+                    c.deleted = True
+            else:
+                l.append(n.symbol.name)
             n.remove()
             n = n.next_term
+
+        for i in reversed(range(len(self.lines))):
+            if self.lines[i].node.deleted:
+                del self.lines[i]
+
         self.skipautolbox = True
         self.reparse(lbox.symbol.ast.children[0], True)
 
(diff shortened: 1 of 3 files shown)


lib/eco/treemanager.py
```diff
@@ -1569,9 +1569,20 @@ def shrink_languagebox(self, lbox, newend):
...
+                for c in list(n.children):
+                    l.append(c.symbol.name)
+                    c.remove()
+                    c.deleted = True
+            else:
+                l.append(n.symbol.name)
             n.remove()
             n = n.next_term
+
+        for i in reversed(range(len(self.lines))):
```
ltratt, line 1582: This looks dodgy: because we `del self.lines[i]` that means we'll skip some items and we could also get an `IndexError`. This should probably be something like:

```python
i = 0
while i < len(self.lines):
  if self.lines[i].node.deleted:
    del self.lines[i]
  else:
    i += 1
```

ltratt:
I'm not quite sure how to parse the title of the third commit "Don't remove-check already deleted boxes".

lib/eco/treemanager.py
ptersilie: I don't think we can skip items or get index errors since I'm running backwards through the list (note the `reversed`). But maybe your variant is more readable?

ltratt: Aha, you're right -- I missed the `reversed`. My version is probably more readable (although, ironically, it is slightly slower), in the sense that it's the more common idiom.

ptersilie:
Fix the title of the third commit. See if that's better.

lib/eco/treemanager.py
ptersilie: So should I change it or leave it?

ltratt: Up to you.

ptersilie: Leave it is then. ;-)

ltratt:
Ready to merge?

ptersilie:
Ready as I'll ever be.

ltratt (approved):

PR #256, opened 2019-07-26
Allow automatic language boxes to split up tokens

ptersilie:
b0e07b1 enables automatic language boxes to be inserted, even if they lead to an existing token needing to be split up. For an example, see the test where a PHP insertion which contains a partial Java comment `//';` is split up  after the `;` to allow the insertion of a language box.

e9dc904 For this to work properly we need to record within the lexer when a node only returned a partial match and forward this information to the automatic language box detector.

### DIFF
--- a/lib/eco/autolboxdetector.py
+++ b/lib/eco/autolboxdetector.py
@@ -78,14 +78,14 @@ class NewAutoLboxDetector(object):
 
         maxdistance = {}
         # find max distance by startnode
-        for start, _, _, dist in valid:
+        for start, _, _, dist, _ in valid:
             if not start in maxdistance or maxdistance[start] < dist:
                 maxdistance[start] = dist
         # remove all startnodes which have less than max distance
         validnew = set()
-        for start, end, lang, dist in valid:
+        for start, end, lang, dist, split in valid:
             if dist >= maxdistance[start]:
-                validnew.add((start, end, lang))
+                validnew.add((start, end, lang, split))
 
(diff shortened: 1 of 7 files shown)


lib/eco/autolboxdetector.py
```diff
@@ -383,6 +390,8 @@ def next_token(self):
             t = self.tokeniter()
             self.last_read = t[3][-1]
             self.last_token_value = t[0]
+            length = sum([len(x.symbol.name) for x in t[3]])
```
ltratt: If this is in a performance critical bit of code, `sum(map(...))` might be faster (if it works)?

ptersilie: Not super critical, but won't hurt.

ptersilie: Nevermind, I just noticed the variable isn't even used. So I can just delete the whole thing.

ltratt:
Is this ready for squashing? If so, please also update the commit messages with information along the lines you've got in the PR description.

ptersilie:
Squashed!

ltratt:
I was sort-of expecting the "fix tests" commit to be squashed, because it's not clear to me which of the other commits broke the tests. Any thoughts?

ptersilie:
It should go into the second commit. I will squash it.

ltratt:
OK, please go ahead.

ptersilie:
Done.

ltratt (approved):

PR #257, opened 2019-08-02
Allow duplicate language definitions

ptersilie:
Previously, the editor would complain if the user defined the same language
in multiple locations. This is especially likely to occur when users define
language compositions and multiple languages embed another language, e.g.
"Python expression". This required the user to then remove the definition
from one of the compositions, only referencing the name of the embeded
language, which is confusing. This change allows the same definition to
occur multiple times without the editor complaining if they are completely
identical. If, however, one of the definition differs while still using
the same name, the editor will as before notify the user about this.

### DIFF
--- a/lib/eco/grammars/grammars.py
+++ b/lib/eco/grammars/grammars.py
@@ -145,6 +145,29 @@ class EcoFile(object):
     def __str__(self):
         return self.name
 
+    def __eq__(self, other):
+        if self.name != other.name:
+            return False
+        if self.base != other.base:
+            return False
+        if self.filename != other.filename:
+            return False
+        if self.alts != other.alts:
+            return False
+        if self.included_langs != other.included_langs:
+            return False
+        if self.extract != other.extract:
+            return False
+        if self.auto_include != other.auto_include:
+            return False
+        if self.auto_exclude != other.auto_exclude:
+            return False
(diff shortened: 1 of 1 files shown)


ptersilie:
I was also thinking about adding more compositions to the editor ready to be used (e.g. the compositions from our experiment for example). Do you think that's a good idea? If so I will do that in a separate PR.

lib/eco/grammars/grammars.py
```diff
@@ -157,6 +180,11 @@ def __hash__(self):
 lang_dict = {}
 
 def add_lang(lang, new=False, sub=False):
+    for l in languages:
```
ltratt: Is this the same as `if lang in languages: return`?

ptersilie: I think you're right!

ptersilie: Fixed.

ltratt:
Please squash.

ptersilie:
Squashed! Any thoughts on my question above?

ltratt:
Oops, sorry! I'm neutral on the additional compositions. We could definitely have too many, but I don't know if we've got to that point yet or not.

ltratt (approved):

PR #261, opened 2019-08-27
Better ranking

ptersilie:
**Not ready for review:** I will do an experiment run to see how this changes our tables and add the observations here.

This PR improves the ranking of automatic languagebox candidates.

Previously, candidates were ranked by the absolute distance they allow parsing to continue. Since this distance had a cut-off of 10 tokens, this gave candidates that occur later in the program an unfair advantage.

We can improve this by taking the candidate that is most likely to parse furthest (i.e. the candidate whose ending boundary reaches the furthest) and check how far it allows parsing to continue (keeping a cut-off of 10 tokens). We then check for each of the remaining candidates if they can parse as far (or further) than that. If they do they are also equally valid candidates. If not, they are worse and are removed from the list of candidates presented to the user.

### DIFF
--- a/lib/eco/autolboxdetector.py
+++ b/lib/eco/autolboxdetector.py
@@ -5,6 +5,7 @@ from incparser.syntaxtable import Shift, Reduce, Goto, Accept
 import config
 
 ws_tokens = ["<ws>", "<return>", "<slcomment>", "<mlcomment>"]
+PARSE_AFTER_TOKENS = 10
 
 def get_lookup(la):
     """Get the lookup symbol of a node. If no such lookup symbol exists use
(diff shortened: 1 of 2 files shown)


lib/eco/autolboxdetector.py
```diff
@@ -116,13 +125,11 @@ def heuristic_line(self, errornode):
                     start = node.next_term
                     result = r.parse(start)
                     if r.possible_ends:
+                        print(r.possible_ends)
```
ltratt: Debugging statement?

lib/eco/autolboxdetector.py
```diff
@@ -177,27 +184,27 @@ def heuristic_history(self, errornode):
...
         if root is not parent:
             # if the parent is already the root we don't need to preparse
             # anything
             ir.preparse(root, parent, version)
         # try parsing lbox + one more non-ws terminal
-        check = ir.parse_single(TextNode(lbox)) and ir.parse_after(end.next_term, split, distance=10)
+        check = ir.parse_single(TextNode(lbox)) and ir.parse_after(end.next_term, split, maxtoks=PARSE_AFTER_TOKENS, maxdist=maxdist)
         self.abs_parse_distance = ir.abs_parse_distance
-        return check
+        if check:
```
ltratt: `return check and (... or ...)`?

ptersilie:
Addressed comments.

ltratt:
Looks fine to me -- I guess we need to wait for draft results first, though, before thinking about squashing.

ptersilie:
Here are the results from the latest run.
Screenshot_2019-08-28_11-13-34

Overall things have improved since submission, though it's difficult to tell how much came from this PR, as there have been various other changes since then, and there will also be some noise, since we cannot run the experiments with the exact same tests as in the submission. It seems though that we now have more "No insertion (Multi)" (i.e. there were multiple insertion options). This matches my intuition as the new ranking now gives earlier insertions a fair chance, compared to those that come later and can thus parse further.

ltratt:
Stack is now noticeably better than Line, which is interesting, although the numbers are close enough that it doesn't make too much difference.

I suggest we squash this and merge it.

ptersilie:
I think Stack only wins overall because Line performs badly on JavaPHP (84%), where functions are so similar that errors often only occur on the second line. For the other benchmarks they seem quite similar with small variances depending on the benchmark.

ptersilie:
Squashed!

ltratt (approved):

ltratt:
@ptersilie Can you try kicking Travis? It might be that it won't work (@vext01 and I had a PR a few days ago which had a similar looking Travis-doesn't-start problem).

ptersilie:
Eco doesn't even show up on Travis at the moment, so there isn't even a button I could press to kick it. I'll investigate.

ptersilie:
I have kicked it, but doesn't look like that helped.

vext01:
The build appears to have run, just github wasn't notified properly:
https://travis-ci.org/softdevteam/eco

ptersilie:
Yes, I've rerun it a couple of times, but it seems the communication between Github and Travis is broken. I'm not sure how to fix it though.

PR #262, opened 2019-09-09
Heuristic runtime

ptersilie:
This PR contains two optimisations that reduce the runtime of the heuristics, removing a worst-case scenario of over 3s for a single keypress in our experiment, and replacing it with a worst-case of 0.13s, which is also the only average going above 0.1s.

However, I noticed that one of the optimisations changes the outcome of the stack heuristic for a specific composition, which is why I disabled it again (see comment). I would prefer to leave it in as ac omment though as it might be useful in the future, but can be convinced otherwise.

### DIFF
--- a/lib/eco/autolboxdetector.py
+++ b/lib/eco/autolboxdetector.py
@@ -113,7 +113,6 @@ class NewAutoLboxDetector(object):
 
     def heuristic_line(self, errornode):
         valid = []
-        pv = self.op.prev_version
         for sub in self.langs:
             lbox = MagicTerminal("<{}>".format(sub))
             node = errornode.prev_term
@@ -129,7 +128,7 @@ class NewAutoLboxDetector(object):
                             if e.lookup == "<ws>" or e.lookup == "<return>":
                                 continue
                             valid.append((start, e, sub, enddist, split, lbox, errornode))
-                if node.lookup == "<return>" or type(node) is BOS:
+                if node.lookup == "<return>" or type(node) is BOS or node.ismultinode():
                     break
                 node = node.prev_term
         return valid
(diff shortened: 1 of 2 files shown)


ltratt:
If we remove that optimisation, does that change the performance by very much?

ptersilie:
As mentioned in the comment, the speedup is quite low (from the top of my head I think it was from 0.15 to 0.12).

ltratt:
The comment doesn't talk about performance, it just mentions that it impacts correctness. IMHO, if there's a correctness implication, we should excise this optimisation entirely, unless, perhaps, we think we can fix it at some point in the not-too-distant future. Especially since it doesn't seem like this makes *very* much difference to performance, I suggest we just remove this optimisation entirely. It seems like the multi-node fix is quite enough on its own!

ptersilie:
> Since the speedup from this is not very significant, we disable this optimisation for the time being.

I didn't mention extact numbers, just that the impact is low. I don't think we can "fix" the correctness. It's not strictly a correctness problem either, we are just reducing our candidate search space. However, this only affects a single benchmark, so our overall insertion stats are barely affected. So it's just a question of whether it's worth sacrificing a single benchmark for a small performance boost.

ltratt:
I suggest we remove those commits and just merge the multi-node commit. It's the easiest thing to explain, if nothing else!

ptersilie:
Ok, I can live with that.

ptersilie:
Please check.

ltratt:
Please squash.

ptersilie:
Sorry, please also check my last commit.

ltratt:
OK, please squash.

ptersilie:
Squashed!

ltratt (approved):

ltratt:
@ptersilie Has Travis died or ...?

ptersilie:
I'll try and kick it. Hope it's not the same problem as last time which we couldn't fix at all.

ltratt:
Travis seems to have run successfully https://travis-ci.org/softdevteam/eco/builds/582721325 so I'm merging, even though GitHub doesn't seem to have picked this up.

PR #269, opened 2019-10-31
Show warning when pygame is not available

ptersilie:
When trying to view the parse tree and pygame isn't installed, show a
warning and disable the menu entry.

### DIFF
--- a/lib/eco/eco.py
+++ b/lib/eco/eco.py
@@ -24,7 +24,7 @@ import subprocess, sys
 
 try:
     import __pypy__
-    print("Warning: PyQt4 is currently not supported in PyPy. Some features thus have been disabled.")
+    print("Warning: PyQt is currently not supported in PyPy. Some features thus have been disabled.")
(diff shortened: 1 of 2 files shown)


lib/eco/eco.py
```diff
@@ -833,6 +834,15 @@ def contextMenu(self, pos):
         self.btReparse([])
 
     def showPygameTree(self):
+        try:
```
ltratt, line 838: This puts more stuff than necessary in the `try` I think? Could we not do something like:

```python
os.environ[...] = ...
try:
    import pygame
except ImportError:
    ...
    return
self.ui.actionPvShow...
```

ptersilie: Fixed. The `self.ui` stuff wasn't needed here so I removed it.

ltratt:
Please squash.

ptersilie:
Squashed!

ltratt (approved):

PR #271, opened 2020-01-11
Addition to docs, fix in JsonManager

ComFreek:
I added further install instructions to the readme such that Ubuntu and WSL users (like me) can directly all necessary packages without diving into the web.

Furthermore, I always got
```
TypeError: the JSON object must be str, not 'bytes'
  File "eco.py", line 1157, in newfile
    etab.set_language(lang, lview.getWhitespace())
  File "/.../eco/lib/eco/editortab.py", line 121, in set_language
    incparser, inclexer = lang.load()
  File "/.../eco/lib/eco/grammars/grammars.py", line 85, in load
    root, language, whitespaces = manager.load(self.filename)[0]
  File "/.../eco/lib/eco/jsonmanager.py", line 52, in load
    main = json.loads(z.read())
  File "/usr/lib/python3.5/json/__init__.py", line 312, in loads
    s.__class__.__name__))
```
on starting Eco. I fixed that in JsonManager, see the small diff.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ distributed under a BSD/MIT license.
 ### Install ###
 At a minimum you will need to install:
 
-* Python 3 https://www.python.org/download/
+* Python 2 **and** 3 https://www.python.org/download/
 * PyQt5 http://www.riverbankcomputing.co.uk/software/pyqt/download5
 * Py http://py.readthedocs.io/en/latest/install.html
 
@@ -26,8 +26,8 @@ If you wish to see visualisations of parse trees, you may optionally install:
 The above dependencies can be installed as follows on command line:
 
 ```
-# Get Python3 if you haven't already
-sudo apt-get install python3
+# Get Python 2 *and* 3 if you haven't already
+sudo apt-get install python python3
 
 python3 -m pip install --user PyQt5 py
 
--- a/README.md
+++ b/README.md
@@ -21,6 +21,24 @@ If you wish to see visualisations of parse trees, you may optionally install:
 * PyDot https://code.google.com/p/pydot/
 * Pygame https://www.pygame.org/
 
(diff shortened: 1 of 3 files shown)


ComFreek:
(The Travis build fails for other reasons than the changes in this PR.)

ptersilie:
Thanks for this PR!

Just so that I can reproduce the error, which JSON file did you try to open?

ComFreek:
@ptersilie To be honest, I don't know. If I remember correctly, this error appeared on start up (`> ./bin/eco`) and directly terminated the application.

You might infer the JSON filename from the stacktrace I posted above. It seems to be some grammar file.

However, the error should be agnostic to the JSON file. It's about bytes vs str objects, which were treated equally in Python 2, I guess.

ptersilie:
I see. I just wanted to check since everything runs fine on my system. But the good news is that your changes to `JsonManager` don't break anything. It might have something to do with Ubuntu or the Windows Subsystem.

Did you try Eco on both Ubuntu and the Windows Subsystem, or WSL only?

ComFreek:
I tried it on the Ubuntu-flavored WSL. (There are also differently flavored WSL variants.)

That you don't get that error really puzzles me. Are you running Eco with Python 3?

ptersilie:
Yes, I'm using Python 3 but I'm running it on Linux. I don't exactly know how the WSL works, but I can imagine that the encoding issue is because I created the json files on my Linux system and you are trying to open them on a Windows system. Anyways, your changes seem to fix it and they don't break anything on my end. So I'm happy to merge them once my travis fix (#276 ) is merged.

While we wait for that, may I ask you to split this PR into two: one with the documentation changes and one with the `JsonManager` fix? I have a few comments on the documentation as well but I will add those once we have split up this PR.

ComFreek:
Split in favor of #277 and #278.

PR #273, opened 2020-01-12
Allow dashes in image filenames.

### LINKED ISSUE #272: Image paths in HTML + Python + SQL file cannot contain dashes `-`
I really wondered why Eco wouldn't accept my image path until I realized the syntax doesn't account for dashes in filenames:
image

ltratt:
The diff is roughly as follows:

  - img:"[a-zA-Z0-9_/\.]+"
  + img:"[a-zA-Z0-9_/\.-]+"

Fixes https://github.com/softdevteam/eco/issues/272.


ltratt:
Do you want me to make this equivalent to `.*`?

ptersilie:
Maybe for the time being that's the easiest solution. The only downside would be that this probably breaks automatic language boxes for compositions with `Image` language boxes.

It seems that Linux allows pretty much anything within a filename but `NUL` and `/`. So `.*` is not too far off anyways.

ltratt:
Try https://github.com/softdevteam/eco/pull/273/commits/785b9f29fdf3344ef97d99ec5cadef821f1c72b6 and see what you think.

ptersilie:
I think that will work. Please squash.

ltratt:
Squashed.

ptersilie (approved):

ComFreek:
Citing from your commit message:

> `+ img:"[/\\].+"`
> i.e. we at least make sure the user uses an absolute path, but don't do anything
beyond that.

FYI this won't work on Windows, though, where absolute paths usually start with a drive letter like `D:\...\...`. One might have luck with the [`\\?\D:\...\...` syntax](https://docs.microsoft.com/de-de/archive/blogs/jeremykuhne/more-on-new-net-path-handling), however that does not even work in PowerShell - at least not in a way I could manage to come up with in the last 3 minutes.

ltratt:
> this won't work on Windows, though, where absolute paths usually start with a drive letter

Good point. I'll have a think about what to do.

ptersilie:
To do this properly we would have to provide different lexing rules for each OS. It seems `.*` is the only thing we can do for now.

PR #279, opened 2020-01-14
Language box fixes

### LINKED ISSUE #274: Incorrect auto lboxes
In `PHP+Python` type this:

```
function f() { }
```

then put your cursor after the `(` and type `$`. `function` is temporarily put in a Python language box, which is a little odd, but maybe OK. Then type `x` (i.e. a function parameter of `$x`) and now `{ }` is an invalid Python language box. Do much more than that and Eco tends to crash!

ptersilie:
This PR contains two issues with language boxes.

- 23a6607 addresses and fixes #274
- ac2f00d addresses a bug when changing the language of a language box while that box contains a parsing error, which causes Eco to crash.

### DIFF
--- a/lib/eco/eco.py
+++ b/lib/eco/eco.py
@@ -714,7 +714,7 @@ class Window(QMainWindow):
         # adjust icon size depending on DPI
         dpi = QApplication.desktop().physicalDpiX()
         if dpi > 96:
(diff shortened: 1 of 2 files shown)


ltratt:
These fix the problem for me too!

ltratt (approved):

